PR #44, opened 2026-02-11
Rebase Weekly

walterchris:


sylv-io:
I don't think it's a good idea to automatically update the submodule every week yet. Our current approach with patch files assumes a specific revision, and we do not pin them. This could result in errors during the build process, whereby the patches do not apply.

We need to discuss how we want to handle recipes with patches. Should we pin them to prevent build errors when the revision changes, or should we not pin them so that we have to update the patch every time it does not apply?

In addition, I would prefer to reduce the frequency to once a month.

walterchris:
Even with one months we are one week behind.

And the weekly rebase is nothing I like to compromise on. This is our core promise and we need to cover this through testing.

sylv-io:
> Even with one months we are one week behind.
> 
> And the weekly rebase is nothing I like to compromise on. This is our core promise and we need to cover this through testing.

Yes, but at the moment, we only test build and don't have a testing infrastructure yet. This could lead to undetected regressions. My proposal would be to add testing first (FirmwareCI) and than add weekly updates.

walterchris:
I'm fine with that.

sinuscosinustan:
Closing in favour of #140

Issue #39, opened 2026-02-10
HPE Proliant Gen11: Minimal thermal and fan control


mynetz:
# Summary

The platform starts with 50% fan speed and no information on temperature. Implement means to control the fans and show/monitor the temperature.
Simple implementation for now. Single temperature zone and all fans to same speed triggered by CPU or other system temperature.

## Tasks
- Implement/access sensors for temperature
- Implement/access CPLD based fan control to set duty cycle
- Feedback from CPLD for fan state (fan good vs not)

## References

- canopybmc/hpe-canopy-dev#5
- canopybmc/hpe-canopy-dev#6
- [openbmc/phosphor-pid-control](https://github.com/openbmc/phosphor-pid-control)

# Note
> This is a tracking item for work that may partially done in different repositories. Items are linked as much as possible. Some repositories are private and stay private since they are used to verify and clean-up vendor (derived) code.

mynetz:
@sylv-io does canopybmc/canopybmc#46 resolve the whole issue?

sylv-io:
> [@sylv-io](https://github.com/sylv-io) does [#46](https://github.com/canopybmc/canopybmc/pull/46) resolve the whole issue?

no this only covers the first part (kernel fan support). I am working on a second PR, which covers the rest (obmc Inventory integration, coretemp support and initial phosphor-pid-control config)

mynetz:
Should be done according to @sylv-io. Only for Intel platforms. AMD related follow-up issue created here: canopybmc/canopybmc#53

PR #46, opened 2026-02-12
HPE Proliant Gen11: add fan control and CPLD host power regulator

sylv-io:
Update the kernel SRCREV to enable fan control and support for the CPLD host power regulator.  Also, add a kernel debug configuration fragment for on-target diagnostics.

The kernel changes include:
- Fix the gxp-fan-ctrl PWM register offset (reads and writes went to the wrong registers) and the dual-motor fan fault detection (only faults in motor 1 were reported. faults in motor 2 were missed).
- Add a CPLD host power supply regulator driver that exposes the power-good state through the regulator framework.
- Rework gxp-fan-ctrl to use the fan-supply regulator for power state detection. Add the standard pwm_enable attribute and hide absent fan channels via is_visible.
- DT bindings and device tree updates for the above changes.

Testing was performed on an HPE ProLiant DL320 Gen11 by removing fans or obstructing the fan rotor to provoke a fault.

Addresses #39


meta-canopy/meta-hpe/meta-gxp/recipes-kernel/linux/linux-canopy_6.18.bbappend
```diff
@@ -1,11 +1,13 @@
 FILESEXTRAPATHS:prepend := "${THISDIR}/${BPN}:"
 
 KBRANCH = "v6.18-hpe-gxp"
-SRCREV = "9c48ad66ee2f0aaf99b948c5e3c0495c22130e3e"
+SRCREV = "c51e9d817f298c1aa1270a9240651f9dd8684aab"
```
pmaione: There is no commit with this hash on branch `v6.18-hpe-gxp`. Bitbake can't fetch.

pmaione: Last commit on `v6.18-hpe-gxp` is `3f52de66badeb20909463d521abbc742df468660`

sylv-io: oh right, i changed it to `3f52de66badeb20909463d521abbc742df468660`

pmaione:
Bulid Ok and tested setting and getting pwm values. Cannot test fan fault on DL360.

Tested on HPE ProLiant DL360 G11 (lab\

PR #47, opened 2026-02-14
meta-gxp: use mmio-mux and i2c-mux-gpmux for i2c bus selection

walterchris:
The xreg node claims a single contiguous region at 0xd1000000 size 0x11c, which covers the i2c mux select registers at offsets 0x84-0x93. This prevents i2c-mux-reg from requesting those addresses, causing all four i2c mux instances to fail probe with -EBUSY.

Split the xreg reg property into two ranges with a gap at 0x84-0x93 so that i2c-mux-reg can claim the mux select registers:
  0xd1000000-0xd1000083 (0x84 bytes)
  0xd1000094-0xd10000ff (0x6c bytes)


walterchris:
Or do you want me to apply the patch directly to the kernel source tree?

meta-canopy/meta-hpe/meta-gxp/recipes-kernel/linux/linux-canopy/0001-ARM-dts-hpe-gxp-split-xreg-reg-range-for-i2c-mux.patch
```diff
@@ -0,0 +1,37 @@
...
+index 1234567..abcdefg 100644
+--- a/arch/arm/boot/dts/hpe/hpe-gxp.dts
++++ b/arch/arm/boot/dts/hpe/hpe-gxp.dts
+@@ -105,7 +105,8 @@
+
+ 		xreg: xreg@d1000000 {
+ 			compatible = "hpe,gxp-xreg", "simple-mfd", "syscon";
+-			reg = <0xd1000000 0x11c>;
++			/* gap at 0x84-0x93 reserved for i2c mux select registers */
++			reg = <0xd1000000 0x84>, <0xd1000094 0x6c>;
```
sylv-io: this will break xreg access to the 2nd half:
`drivers/mfd/syscon.c:113`
```
syscon_config.max_register = res_size - reg_io_width
```
technically it still works, because we do not use any reg from the 2nd half yet.
the proper solution would be to use `mmio-mux` + `i2c-mux` where `mmio-mux` is defined as child node of `xreg`. by changing `i2c-mux-reg` to `i2c-mux` it should be able to use mux-reg without exclusive claim to it.

meta-canopy/meta-hpe/meta-gxp/recipes-kernel/linux/linux-stable/debug.cfg
```diff
@@ -1,2 +1,5 @@
 CONFIG_DEBUG_FS=y
 CONFIG_DYNAMIC_DEBUG=y
+CONFIG_MULTIPLEXER=y
+CONFIG_MUX_MMIO=y
+CONFIG_I2C_MUX_GPMUX=y
```
sylv-io: `CONFIG_MULTIPLEXER`, `CONFIG_MUX_MMIO`, and `CONFIG_I2C_MUX_GPMUX` are functional configs, not debug configs. move it to the defconfig.

sylv-io (approved):

sylv-io (changes requested):
I can confirm that this fix works on the DL320. i don't see -EBUSY anymore and the i2cmux bus seems to work. 👍 

nit: The commit message refers to your old approach. Update it so that it describes the fix independently.
Maybe also update the title of this PR to match the changes.

PR #48, opened 2026-02-14
meta-hpe: expose baseboard PCA VPD via entity-manager

walterchris:
Expose baseboard-level Vital Product Data (PCA serial number and PCA part number) on D-Bus at /xyz/openbmc_project/MachineContext/Baseboard.

U-Boot's ft_board_setup() populates a /baseboard subnode in the Linux device tree with PCA serial-number and part-number properties. This change patches the entity-manager devicetree-vpd-parser to read those properties and expose them via the Inventory.Decorator.Asset D-Bus interface.


walterchris:
Same here - should I rather commit it to a fork for now - or do we want to keep the patch file?

sylv-io (changes requested):
https://github.com/canopybmc/u-boot/commit/32695be70ae44fc4f9e953f4bd312ad5d86468a6 is missing. add it as additional u-boot patch.

PR #49, opened 2026-02-14
Hpe/feature/psu enablement

walterchris:


meta-canopy/meta-hpe/meta-common/recipes-phosphor/configuration/entity-manager/dl360g11_baseboard.json
sylv-io, line 1: This is missing for all other ProLiant Gen11 targets.

sylv-io (approved):
Apart from the missing association for all non-DL360 targets, this looks good to me.
One thing that caught my attention is that the inventory only lists available PSUs.  e.g.: when I remove PSU2 from the DL320, PSU1 is listed as enabled, but PSU2 is not listed as absent at all.
We could address this either in this PR or in a separate one. Your decision :)

sylv-io:
> One thing that caught my attention is that the inventory only lists available PSUs. e.g.: when I remove PSU2 from the DL320, PSU1 is listed as enabled, but PSU2 is not listed as absent at all. We could address this either in this PR or in a separate one. Your decision :)

Additionally, hotplugging is not supported. Attempting to connect a PSU while the system is running does not result in the PSU being listed as available.

walterchris:
Created #63 to track this

sylv-io:
> Apart from the missing association for all non-DL360 targets, this looks good to me.

you missed this comment. :sweat_smile:

meta-canopy/meta-hpe/meta-common/recipes-phosphor/configuration/entity-manager/dl360g11_baseboard.json
sylv-io: still missing. should we add this in a separate PR?

sylv-io:
@walterchris: You merged this into `hpe/feature/VPDPCA` instead of `hpe/main`. Was this intentional?

walterchris:
Jesus Christ.. nope.

PR #50, opened 2026-02-15
meta-hpe: add total system power virtual sensor

walterchris:
Add phosphor-virtual-sensor to the image with a configuration that computes total system power consumption by summing PSU1 and PSU2 input power readings (PINPUT_PSU1 + PINPUT_PSU2).

The expression handles PSU absence gracefully: if only one PSU reports data, total_power reflects that single PSU; if neither reports, the sensor reads NULL.

This is similar to what HPE has enabled in their openBMC version


pmaione:
@walterchris I did this on another project, it also applies here: [Redacted internal link]

This allows bmcweb to filter sensors by purpose, so that `total_power` sensor show on the chassis environment metrics for `PowerWatts`.

walterchris:
So right now I don't carry around any changes to the `virtual-sensor` - however you are bringing in some change to that. Is that something you can upstream to openbmc and see what's the feedback on this?

pmaione:
I'll upstream this.

walterchris:
Can you add a ticket on this to follow-up on this again?

pmaione:
[Redacted internal project link]

PR #64, opened 2026-02-19
Hpe/feature/psu enablement

walterchris:


meta-canopy/meta-hpe/meta-common/recipes-phosphor/configuration/entity-manager/dl360g11_baseboard.json
sylv-io, line 1: This is missing for all other ProLiant Gen11 targets.

sylv-io (changes requested):

sylv-io (approved):

Issue #63, opened 2026-02-19
HPE Proliant Gen11: Support PSU hotplug


walterchris:
# Summary

Currently the PSUs don't support hot plug. We try to read the FRU and if it's not available we don't add it to the inventory.

PR #65, opened 2026-02-21
meta-hpe: add PSU presence detection

walterchris:
Add GPIODeviceDetect entries (INST_PSU1, INST_PSU2) to all Gen11 baseboard configs for GPIO-based PSU presence detection.

Split hpe_psu.json from a single templated config into two explicit PSU entries (PSU1 at bus 11/addr 88, PSU2 at bus 10/addr 89) with per-slot ADDRESS-filtered FRU probes, sensor thresholds, and named sensor labels.


walterchris:
We should wait to merge this until we rebase. The patch has been merged upstream - thus we can drop it if wen move the `SRCREV` forward.

meta-canopy/meta-hpe/meta-common/recipes-phosphor/configuration/entity-manager/dl560g11_baseboard.json
```diff
@@ -337,6 +337,18 @@
...
+            "PresencePinNames": ["INST_PSU1"],
+            "PresencePinValues": [1],
+            "Type": "GPIODeviceDetect"
+        },
+        {
+            "Name": "PSU2Detect",
+            "PresencePinNames": ["INST_PSU2"],
+            "PresencePinValues": [1],
+            "Type": "GPIODeviceDetect"
+        },
```
sinuscosinustan, line 351: Doesn't have the DL560G11 four PSUs?

sylv-io (dismissed):
This PR needs to be rebased. I could test and verify its functionality locally on the DL320 tomorrow.

mynetz:
Can we get this in soonish? Was that tied to in person testing?

sinuscosinustan:
@walterchris Could you please test this on-site on our test hardware? If you're too busy right now then please let me know so that we can check who could take over the task.

TheGregggg:
Rebased and removed the patch because it's upstream now.
Tested on the DL365. It's not working. On the inventory page, the removed PSU remains as 'OK' and does not change to 'missing' or anything else. 

What should be the correct behavior ? 

I can modify the probe on the hpe_psu.json EM, so that it matches with the FRU device AND PSU detection:

```
"Probe": [
    "xyz.openbmc_project.FruDevice({'PRODUCT_PRODUCT_NAME': 'HPE POWER SUPPLY*', 'ADDRESS': 88})",
    "AND",
    "xyz.openbmc_project.Inventory.Source.DevicePresence({'Name': 'PSU1Detect'})"
]
```
This will remove the PSU from the inventory page and all the related sensors from the sensors page. I don't know if this would be the intended behavior.

No redfish logs appeared. And even with the proposed solution above, no logs will be emitted. It's another issue to tackle if log emission on PSU presence changes is intended behavior.

walterchris:
What happens if you boot the BMC with one PSU in? Is the second PSU shown as missing/error ? Or does it not show up in the entity manager at all?

TheGregggg:
The second PSU doesn't show up at all. There is only the first PSU in the inventory and all the related sensors in the sensors page

TheGregggg:
> I can modify the probe on the hpe_psu.json EM, so that it matches with the FRU device AND PSU detection:
> 
> ```
> "Probe": [
>     "xyz.openbmc_project.FruDevice({'PRODUCT_PRODUCT_NAME': 'HPE POWER SUPPLY*', 'ADDRESS': 88})",
>     "AND",
>     "xyz.openbmc_project.Inventory.Source.DevicePresence({'Name': 'PSU1Detect'})"
> ]
> ```
> 
> This will remove the PSU from the inventory page and all the related sensors from the sensors page. I don't know if this would be the intended behavior.

I have tested this. It's working as intended if the BMC boots with the PSU plugged in. The FruDevice is not refreshed when the PSU is plugged in, so this doesn't work if the PSU is not present when the BMC boots. 

I can work on upstreaming a change to gpio-monitor to refresh fru devices on a given bus, on presence detection.
Alex thinks it's a good idea. 

From my own research, there is nothing providing this functionality in openbmc without creating our own EM reactor. I think it makes more sense to have gpio-monitor provide this functionality. 

This would add: 
```
"ReScanBuses": [20],
```
to the GPIODeviceDetect entries.

This will also fix #63.

What do you think about this ?

I don't have any solution for the redfish log entry for now.

walterchris:
@pointbazaar any feedback on the `gpio-monitor` changes?

Issue #77, opened 2026-03-05
Base toolchain fails to build with GCC 15.2.1


sinuscosinustan:
## Issue description

Arch Linux provides GCC `15.2.1 (20260204)` which is being used by `bitbake` to build the base toolchain.  
Due to re-declarations by GCC 15 for certain macros like `memchr`, the toolchain fails to build `m4-native` and other packages:

`m4-native`
```
| In file included from /usr/include/features.h:540,
|                  from /usr/include/errno.h:25,
|                  from ../../m4-1.4.20/lib/closeout.c:23:
| ./string.h:777:20: error: expected identifier or '(' before '_Generic'
|   777 | _GL_EXTERN_C void *memchr (const void *__s, int __c, size_t __n)
|       |                    ^~~~~~
| In file included from /usr/include/features.h:540,
|                  from /usr/include/bits/libc-header-start.h:33,
|                  from /usr/include/stdio.h:28,
|                  from ./stdio.h:55,
|                  from ../../m4-1.4.20/lib/clean-temp.h:26,
|                  from ../../m4-1.4.20/lib/clean-temp.c:22:
| ./stdlib.h:827:20: error: expected identifier or '(' before '_Generic'
|   827 | _GL_EXTERN_C void *bsearch (const void *__key,
|       |                    ^~~~~~~
| ./string.h:777:20: error: expected identifier or '(' before '_Generic'
|   777 | _GL_EXTERN_C void *memchr (const void *__s, int __c, size_t __n)
|       |                    ^~~~~~
| ./string.h:777:20: error: expected identifier or '(' before '_Generic'
|   777 | _GL_EXTERN_C void *memchr (const void *__s, int __c, size_t __n)
|       |                    ^~~~~~
| ./stdlib.h:827:20: error: expected identifier or '(' before '_Generic'
|   827 | _GL_EXTERN_C void *bsearch (const void *__key,
|       |                    ^~~~~~~
| ./stdlib.h:827:20: error: expected identifier or '(' before '_Generic'
|   827 | _GL_EXTERN_C void *bsearch (const void *__key,
|       |                    ^~~~~~~
| make[3]: *** [Makefile:4482: libm4_a-chdir-long.o] Error 1
| make[3]: *** [Makefile:4454: libm4_a-canonicalize.o] Error 1
| make[3]: *** [Makefile:4594: libm4_a-closeout.o] Error 1
| make[3]: *** [Makefile:4580: libm4_a-closein.o] Error 1
[...]

PR #78, opened 2026-03-06
doc: introduce bug template and update feature template

mynetz:
This adds a new bug template following the example from canopybmc/canopybmc#77.

Also the title of the feature template has been updated.


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,36 @@
...
+about: Create a report to help us improve
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+# Issue Summary
+
+Descriptive summary of the feature
```
sinuscosinustan: I'd suggest to make this as a Markdown comment (and all the other ones)

```suggestion
<!-- Descriptive summary of the feature -->
```

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,36 @@
...
+## Build environment
+
+```
+[Host: <name>]
+OS:
+	<cat /etc/os-release (if available)>
+		_OR_
+	<Ubuntu x.y|Fedora x|Arch>
+Python: <version>
+GCC: <version>
```
sinuscosinustan, line 35: It _may_ also make sense to add the build target (which I completely forgot because we only have one at this time).


```suggestion
GCC: <version>
Selected build target: <target>
```

.github/ISSUE_TEMPLATE/bug_report.md
mynetz: Done in commit b3725920e05a2934674f70f737a2c25bc548e2a2

sinuscosinustan (approved):
LGTM. Thanks for implementing this!

PR #90, opened 2026-03-11
meta-hpe: add thermal zone configurations

### LINKED ISSUE #57: HPE Proliant Gen 11: Fine tune Thermal Zones
# Summary

#52 introduces thermal control for the HPE Proliant Series - we need to fine-tune the configurations and do testing around it.

sinuscosinustan:


meta-canopy/meta-hpe/meta-common/recipes-phosphor/configuration/entity-manager/dl320g11_baseboard.json
```diff
@@ -134,7 +134,7 @@
             "Name": "Fan 1-7",
             "NegativeHysteresis": 0,
             "OutLimitMax": 100,
-            "OutLimitMin": 30,
+            "OutLimitMin": 0,
```
walterchris: Do we really want to allow this? Or should we use a lower limit of 10 or so?

walterchris: This also accounts for the zones - I think it would be safe to set a minimum output level that is not zero - just in case?

sinuscosinustan: Leftover from debugging :see_no_evil: Thanks! Will set it to 10% minimum.

walterchris (approved):

PR #105, opened 2026-03-18
Hpe/feat/qemu

walterchris:
This PR is introducing two things:
1. It adds a canopy-qemu target which uses the ast2600 as a baseline and picks the meta-canopy layer on top
2. It moves some of the tests in main from the qemu target in order to reduce the load on the HW


walterchris:
All checks  now pass - @sinuscosinustan - ready for review :)

sylv-io (approved):
lgtm

meta-canopy/conf/templates/default/bblayers.conf.sample
```diff
@@ -0,0 +1,18 @@
...
+
+BBPATH = "${TOPDIR}"
+BBFILES ?= ""
+
+BBLAYERS ?= " \
+  ##OEROOT##/meta-canopy \
+  ##OEROOT##/openbmc/meta \
+  ##OEROOT##/openbmc/meta-openembedded/meta-oe \
+  ##OEROOT##/openbmc/meta-openembedded/meta-networking \
+  ##OEROOT##/openbmc/meta-openembedded/meta-perl \
```
sinuscosinustan, line 13: I don't think we have anything perl-related in use

meta-canopy/conf/templates/default/bblayers.conf.sample
```diff
@@ -0,0 +1,18 @@
...
+  ##OEROOT##/openbmc/meta-openembedded/meta-python \
+  ##OEROOT##/openbmc/meta-openembedded/meta-webserver \
```
sinuscosinustan, line 15: Unnecessary. `bmcweb` provides its own HTTP server and we don't use any recipes from there

meta-canopy/conf/templates/default/local.conf.sample
```diff
@@ -0,0 +1,17 @@
+MACHINE ??= "canopy-qemu"
+DISTRO ?= "openbmc-phosphor"
+PACKAGE_CLASSES ?= "package_ipk"
+SANITY_TESTED_DISTROS:append ?= " *"
+EXTRA_IMAGE_FEATURES ?= "allow-root-login"
+USER_CLASSES ?= "buildstats"
+PATCHRESOLVE = "noop"
```
sinuscosinustan, line 7: You could add `QB_GRAPHICS = "-nographic"` to the local conf which would add the nographic parameter to `runqemu` automatically.

meta-canopy/conf/machine/canopy-qemu.conf
```diff
@@ -0,0 +1,24 @@
...
+require conf/machine/include/obmc-bsp-common.inc
+include conf/machine/include/obmc-evb-common.inc
+
+SERIAL_CONSOLES = "115200;ttyS4"
+
+# AST2600 EVB default: 64 MB SPI flash
+FLASH_SIZE = "65536"
+
+# Use the Aspeed kernel, not the Canopy GXP kernel
+PREFERRED_PROVIDER_virtual/kernel = "linux-aspeed"
```
sinuscosinustan, line 24: redundant. the GXP kernel is being set via `conf/machine/include/hpe_soc.inc` which is only available in the GXP layer

sinuscosinustan (commented):
A bit late to the show but here we are 🙈

PR #111, opened 2026-03-20
meta-hpe: smbios-mdr: skip broken type 0 tables

### LINKED ISSUE #110: HPE DL365 Gen11: Multiple SMBIOS Type 0 tables cause BIOS version to be empty
# Issue Summary

The active BIOS version is not being reported by `smbios-mdrv2` on the DL365 Gen11 and other AMD SP5 systems after receiving the SMBIOS tables from the UEFI via CHIF:

```
root@hpe-proliant-g11:~# busctl introspect xyz.openbmc_project.Smbios.MDR_V2 /xyz/openbmc_project/inventory/system/chassis/motherboard/bios
NAME                                             TYPE      SIGNATURE RESULT/VALUE                           FLAGS
org.freedesktop.DBus.Introspectable              interface -         -                                      -
.Introspect                                      method    -         s                                      -
org.freedesktop.DBus.Peer                        interface -         -                                      -
.GetMachineId                                    method    -         s                                      -
.Ping                                            method    -         -                                      -
org.freedesktop.DBus.Properties                  interface -         -                                      -
.Get                                             method    ss        v                                      -
.GetAll                                          method    s         a{sv}                                  -
.Set                                             method    ssv       -                                      -
.PropertiesChanged                               signal    sa{sv}as  -                                      -
xyz.openbmc_project.Common.UUID                  interface -         -                                      -
.UUID                                            property  s         "00000000-0000-0000-0000-000000000000" emits-change writable
xyz.openbmc_project.Inventory.Decorator.Revision interface -         -                                      -
.Version                                         property  s         ""                                     emits-change writable
```

The reason is that the SMBIOS dump has two Type 0 tables where the first one is empty / only 5 bytes:
```
$ dmidecode --from-dump smbios2-dl365-g11-edited -t bios
# dmidecode 3.7
Reading SMBIOS/DMI data from file smbios2.
SMBIOS 3.3.0 present.

Handle 0x0000, DMI type 0, 5 bytes
Platform Firmware Information

Handle 0x0009, DMI type 0, 26 bytes
Platform Firmware Information
        Vendor: HPE
        Version: 2.90
        Release Date: 01/09/2026
        Address: 0xF0000
        Runtime Size: 64 KiB
        ROM Size: 64 MiB
        Characteristics:
                PCI is supported
                PNP is supported
                Firmware is upgradeable
                Firmware shadowing is allowed
                ESCD support is available
                Boot from CD is supported
                Selectable boot is supported
                EDD is supported
                5.25"/360 kB floppy services are supported (int 13h)
                5.25"/1.2 MB floppy services are supported (int 13h)
                3.5"/720 kB floppy services are supported (int 13h)
                Print screen service is supported (int 5h)
                8042 keyboard services are supported (int 9h)
                Serial services are supported (int 14h)
                Printer services are supported (int 17h)
                CGA/mono video services are supported (int 10h)
                ACPI is supported
                USB legacy is supported
                BIOS boot specification is supported
                Function key-initiated network boot is supported
                Targeted content distribution is supported
                UEFI is supported
        Platform Firmware Revision: 2.90
        Embedded Controller Firmware Revision: 0.0
```

For comparison, this is the SMBIOS dump on a DL360 Gen11:
```
$ dmidecode --from-dump smbios2-dl360-g11-edited -t bios
# dmidecode 3.7
Reading SMBIOS/DMI data from file smbios2-dl360.
SMBIOS 3.3.0 present.

Handle 0x0004, DMI type 0, 26 bytes
Platform Firmware Information
        Vendor: HPE
        Version: 2.50
        Release Date: 04/22/2025
        Address: 0xF0000
        Runtime Size: 64 KiB
        ROM Size: 64 MiB
        Characteristics:
                PCI is supported
                PNP is supported
                Firmware is upgradeable
                Firmware shadowing is allowed
                ESCD support is available
                Boot from CD is supported
                Selectable boot is supported
                EDD is supported
                5.25"/360 kB floppy services are supported (int 13h)
                5.25"/1.2 MB floppy services are supported (int 13h)
                3.5"/720 kB floppy services are supported (int 13h)
                Print screen service is supported (int 5h)
                8042 keyboard services are supported (int 9h)
                Serial services are supported (int 14h)
                Printer services are supported (int 17h)
                CGA/mono video services are supported (int 10h)
                ACPI is supported
                USB legacy is supported
                BIOS boot specification is supported
                Function key-initiated network boot is supported
                Targeted content distribution is supported
                UEFI is supported
        Platform Firmware Revision: 2.50
        Embedded Controller Firmware Revision: 0.0
```

## Expected behavior

`smbios-mdrv2` reports a warning and jumps to the next Type 0 table.

## Current behavior

Parser only reads the first Type 0 table and reports the non-existing version.

## Workaround

Not yet

## Build environment

```
Selected build target: hpe-proliant-g11
```

sinuscosinustan:
Some UEFI firmwares tend to send more than one Type 0 table, though the SMBIOS spec requires only one BIOS table to be included, as stated in section 6.2, Table 4 "Required structures and data". On HPE AMD SP5 systems (specifically, the DL365 and DL385 Gen11), there are two Type 0 tables, and the first one does not have any information.

Workaround the broken SMBIOS dump by skipping all Type 0 tables that are less than 7 bytes. A fix in the firmware must be done by HPE.

Tested: Ensure that the BIOS version is being reported on the DL365
        Gen11 and DL360 Gen11 (sanity check).

Link: https://www.dmtf.org/sites/default/files/standards/documents/DSP0134_3.4.0.pdf

Closes #110

### DIFF
--- a/meta-canopy/meta-hpe/meta-common/recipes-phosphor/smbios/smbios-mdr/0001-Skip-undersized-SMBIOS-entries-in-getSMBIOSTypePtr.patch
+++ b/meta-canopy/meta-hpe/meta-common/recipes-phosphor/smbios/smbios-mdr/0001-Skip-undersized-SMBIOS-entries-in-getSMBIOSTypePtr.patch
@@ -1,0 +1,103 @@
+From ad3c12a864b4e95a50e8e3f17020bd6dbea8ece3 Mon Sep 17 00:00:00 2001
+From: Marcello Sylvester Bauer <marcello.bauer@9elements.com>
+Date: Tue, 24 Mar 2026 13:18:18 +0100
+Subject: [PATCH] Skip undersized SMBIOS entries in getSMBIOSTypePtr
+
+getSMBIOSTypePtr aborts the entire search when a matching entry
+has length < size. Change to skip the entry instead.
+
+Upstream-Status: Pending
+Change-Id: Id4131bd1919602c22d515b98e1fc01330ae94024
+Signed-off-by: Marcello Sylvester Bauer <marcello.bauer@9elements.com>
+Signed-off-by: Tan Siewert <tan.siewert@9elements.com>
+---
+ include/smbios_mdrv2.hpp | 31 +++++++++++++++++--------------
+ src/system.cpp           |  4 ++--
+ 2 files changed, 19 insertions(+), 16 deletions(-)
+
+diff --git a/include/smbios_mdrv2.hpp b/include/smbios_mdrv2.hpp
+index 1bf7733..46b9584 100644
+--- a/include/smbios_mdrv2.hpp
++++ b/include/smbios_mdrv2.hpp
+@@ -17,6 +17,7 @@
+ #pragma once
+ 
+ #include <phosphor-logging/elog-errors.hpp>
++#include <phosphor-logging/lg2.hpp>
+ 
+ #include <array>
+ #include <string>
+@@ -241,32 +242,34 @@ static inline uint8_t* getSMBIOSTypePtr(uint8_t* smbiosDataIn, uint8_t typeId,
+     {
+         return nullptr;
+     }
+-    char* smbiosData = reinterpret_cast<char*>(smbiosDataIn);
++    uint8_t* smbiosData = smbiosDataIn;
+     while ((*smbiosData != '\0') || (*(smbiosData + 1) != '\0'))
+     {
+         uint32_t len = *(smbiosData + 1);
+         if (*smbiosData != typeId)
+         {
+-            smbiosData += len;
+-            while ((*smbiosData != '\0') || (*(smbiosData + 1) != '\0'))
++            smbiosData = smbiosNextPtr(smbiosData);
++            if (smbiosData == nullptr)
+             {
+-                smbiosData++;
+-                len++;
+-                if (len >= mdrSMBIOSSize) // To avoid endless loop
+-                {
+-                    return nullptr;
+-                }
++                return nullptr;
+             }
+-            smbiosData += separateLen;
+             continue;
+         }
+         if (len < size)
+         {
+-            phosphor::logging::log<phosphor::logging::level::ERR>(
+-                "Record size mismatch!");
+-            return nullptr;
++            // Skip undersized entry, same as type-mismatch path above
++            lg2::warning(
++                "SMBIOS type {TYPE} entry too short ({LEN} < {MIN}), skipping",
++                "TYPE", static_cast<uint8_t>(*smbiosData), "LEN", len, "MIN",
++                size);
++            smbiosData = smbiosNextPtr(smbiosData);
++            if (smbiosData == nullptr)
++            {
++                return nullptr;
++            }
++            continue;
+         }
+-        return reinterpret_cast<uint8_t*>(smbiosData);
++        return smbiosData;
+     }
+     return nullptr;
+ }
+diff --git a/src/system.cpp b/src/system.cpp
+index b3e58bd..fa6f1d4 100644
+--- a/src/system.cpp
++++ b/src/system.cpp
+@@ -37,7 +37,7 @@ namespace smbios
+ std::string System::uuid(std::string /* value */)
+ {
+     uint8_t* dataIn = storage;
+-    dataIn = getSMBIOSTypePtr(dataIn, systemType);
++    dataIn = getSMBIOSTypePtr(dataIn, systemType, sizeof(SystemInfo));
+     if (dataIn != nullptr)
+     {
+         auto systemInfo = reinterpret_cast<struct SystemInfo*>(dataIn);
+@@ -124,7 +124,7 @@ std::string System::version(std::string /* value */)
+ {
+     std::string result = "No BIOS Version";
+     uint8_t* dataIn = storage;
+-    dataIn = getSMBIOSTypePtr(dataIn, biosType);
++    dataIn = getSMBIOSTypePtr(dataIn, biosType, sizeof(BIOSInfo));
+     if (dataIn != nullptr)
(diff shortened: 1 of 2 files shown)


meta-canopy/meta-hpe/meta-common/recipes-phosphor/smbios/smbios-mdr/0001-smbios-mdr-skip-broken-type-0-tables.patch
```diff
@@ -0,0 +1,44 @@
...
+     {
+         uint32_t len = *(smbiosData + 1);
+-        if (*smbiosData != typeId)
++        // Some HPE systems report two type 0 tables, where one of them is
++        // 5 bytes and the other one 26 bytes. Usually, the 5 byte table is
++        // empty and is being parsed first, and the other one contains the
++        // real data.
++        // If the length of a Type 0 table is less than 7 bytes, then we
++        // skip it.
++        if (*smbiosData != typeId || (typeId == 0 && len < 7))
```
mynetz: Out of curiosity: If the empty/wrong table is 5 bytes, why does the check require `<7`? I assume the 6th byte is a `\0`.

I guess that is fine enough for the HPE case. A brief skim through the spec shows this for BIOS Information (type 0) tables - Length field:
> 12h + number of BIOS Characteristics Extension Bytes. If no Extension Bytes are used the Length is 12h. For version 2.1 and 2.2 implementations, the length is 13h because one extension byte is defined. For version 2.3 and later implementations, the length is at least 14h because two extension bytes are defined. For version 2.4 to 3.0, implementations, the length is at least 18h because bytes 14-17h are defined. For version 3.1 and later implementations, the length is at least 1Ah because bytes 14-19h are defined.

Prompting two possible more standard aligned solutions:
1. Expect at least 18 bytes
2. Or be even more complex and check for the version and expect 18, 19, 24 or 26 bytes (note: converted to base 10 from the text above)

I guess for what we intend here the first would be enough. Didn't check the code, isn't there a validation function for the tables?

sinuscosinustan: > why does the check require <7? I assume the 6th byte is a \0.

Correct, I took the null terminator into account there which is in the `len`.

> Expect at least 18 bytes

This definitely sounds more reasonable than the "check for only 6 bytes or less" thing I did.

> Didn't check the code, isn't there a validation function for the tables?

Not that I've found one. Theoretically, we could leave `getSMBIOSTypePtr` untouched and do the validation in `System::version` instead (by checking the length and jump to the next table pointer if it is invalid), which may be cleaner than the hack I did.

walterchris:
This feels somewhat hackerish. How about a general parsing of multiple entries and augment or overwrite them if already existent?

sylv-io:
I don't think we need to create a workaround that applies to the exact conditions of the duplicate Type 0 entry on the HPE DL365 and DL385 hosts.

Both the Linux kernel and dmidecode parser skip invalid table entries instead of aborting the parsing process. I suggest you do the same here.

Here is an example patch: 
[0001-Skip-undersized-SMBIOS-entries-in-getSMBIOSTypePtr.patch](https://github.com/user-attachments/files/26344712/0001-Skip-undersized-SMBIOS-entries-in-getSMBIOSTypePtr.patch)

walterchris:
Yeah - that looks better to me.

sinuscosinustan:
I definitely like @sylv-io's implementation way more. Will push this one instead.

meta-canopy/meta-hpe/meta-common/recipes-phosphor/smbios/smbios-mdr/0001-Skip-undersized-SMBIOS-entries-in-getSMBIOSTypePtr.patch
```diff
@@ -0,0 +1,75 @@
...
+-            phosphor::logging::log<phosphor::logging::level::ERR>(
+-                "Record size mismatch!");
+-            return nullptr;
++            // Skip undersized entry, same as type-mismatch path above
++            lg2::warning(
++                "SMBIOS type {TYPE} entry too short ({LEN} < {MIN}), skipping",
++                "TYPE", static_cast<uint8_t>(*smbiosData), "LEN", len, "MIN",
++                size);
++            smbiosData += len;
++            while ((*smbiosData != '\0') || (*(smbiosData + 1) != '\0'))
```
walterchris: Can you use `smbiosNextPtr` here?

sylv-io: This does indeed seem to be the correct method for skipping entries.  I just don't understand why it wasn't used in the `smbiosData!=typeId` condition before.

both `smbiosNextPtr` and `getSMBIOSTypePtr` where introduced by the same author on the same commit `43c6a1d` . 😅

sylv-io: I think we should use `smbiosNextPtr` in both cases. 👍

sinuscosinustan: done ^^

sylv-io (approved):
lgtm

PR #114, opened 2026-03-25
meta-hpe: improve KVM keyboard performance

walterchris:
We are currently seeing a lag and duplicate characters while using KVM. This is an attempt to soften this a bit. This commit introduces three changes:

- Reduce EAGAIN retry to 1ms
- Reduce input polling loop to 1ms
- Add TCP_NODELAY to bmcweb


meta-canopy/meta-hpe/meta-common/recipes-graphics/obmc-ikvm/obmc-ikvm/0002-ikvm-reduce-HID-write-retry-delay-from-10ms-to-1ms.patch
```diff
@@ -0,0 +1,38 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Christian Walter <christian.walter@9elements.com>
+Date: Tue, 25 Mar 2026 00:00:00 +0000
+Subject: [PATCH] ikvm: reduce HID write retry delay from 10ms to 5ms
```
sinuscosinustan: You're saying that the delay is being reduced to 5ms, but the code says 1ms?

sinuscosinustan (changes requested):

meta-canopy/meta-hpe/meta-common/recipes-graphics/obmc-ikvm/obmc-ikvm/0002-ikvm-reduce-HID-write-retry-delay-from-10ms-to-1ms.patch
walterchris: Right Right Right.. I played around with the values a bit and reworked them later. Forgot to update the patch commit message and comment.

meta-canopy/meta-hpe/meta-common/recipes-graphics/obmc-ikvm/obmc-ikvm/0003-ikvm-decouple-input-processing-from-video-frame-rate.patch
```diff
@@ -0,0 +1,46 @@
...
++     * keyboard/mouse latency from ~33ms to ~1ms. */
++    constexpr long int inputPollTime = 1000; /* 1ms per poll */
++    long int remaining = processTime;
++
++    while (remaining > 0)
++    {
++        long int pollTime = (remaining < inputPollTime) ? remaining : inputPollTime;
++        rfbProcessEvents(server, pollTime);
++        remaining -= pollTime;
++    }
```
sinuscosinustan: The problem that I see with this implementation is that it'll increase the required CPU cycles by an insane amount and only addresses the symptom, but not the issue.

From what I could find out, it seems that this just polls more often but is still lock-stepped, meaning it'll still only dispatch input once per frame cycle, just with more `select()` calls per cycle. The server thread and video thread are synchronised via condition variables in the `Manager` (`waitServer`/`waitVideo`), so no matter of how often we run `rfbProcessEvents`, the whole pipeline still runs in lockstep: `server.run() -> Manager::setServerDone -> Manager::waitVideo -> Video::getFrame -> Server::sendFrame -> Manager::setVideoDone -> Manager::waitServer -> repeat`.

The actual fix, I think, needs to decouple the two threads entirely. If the server thread can call `rfbProcessEvents` independently of the video capture loop, input events get dispatched as soon as they arrive, without waiting for the frame cycle. The mutex only needs to protect the shared libvncserver state (`sendFrame`, `resize`), not the entire iteration.

sinuscosinustan (commented):
The other changes look fine to me.

meta-canopy/meta-hpe/meta-common/recipes-graphics/obmc-ikvm/obmc-ikvm/0001-ikvm-decouple-input-thread-from-video-frame-rate.patch
```diff
@@ -0,0 +1,143 @@
...
++    it = rfbGetClientIterator(server);
++    while ((cl = rfbClientIteratorNext(it)))
++    {
++        rfbUpdateClient(cl);
++    }
++    rfbReleaseClientIterator(it);
++
++    /* Maintain the original frame-rate timing budget so the lock-step
++     * cycle does not spin faster than intended. */
++    usleep(processTime);
```
sinuscosinustan: Unconditional sleep. `rfbProcessEvents` returns early by itself when it's ready. And since we still are locked with `setServerDone()` and `waitVideo()`, it'll just add more latency to the whole process.

meta-canopy/meta-hpe/meta-common/recipes-graphics/obmc-ikvm/obmc-ikvm/0001-ikvm-decouple-input-thread-from-video-frame-rate.patch
```diff
@@ -0,0 +1,143 @@
...
+devices without hardware JPEG), the server thread is parked in
+waitVideo() for the entire capture duration. This means key-release
+HID reports are delayed by up to 500ms, long enough for the host
+OS key-repeat to fire and produce duplicate characters.
+
+Fix this by adding a dedicated input thread that calls
+rfbCheckFds() in a free-running 5ms loop, completely independent
+of the video frame cycle. The server thread no longer calls
+rfbProcessEvents() (which internally calls rfbCheckFds) and
+instead iterates clients with rfbUpdateClient() for framebuffer
```
sinuscosinustan: No. `rfbUpdateClient()` also dispatches other events (`ptrAddEvent`). So the server thread still processes some input, and the input thread does not exclusively handle it.

meta-canopy/meta-hpe/meta-common/recipes-graphics/obmc-ikvm/obmc-ikvm/0001-ikvm-decouple-input-thread-from-video-frame-rate.patch
```diff
@@ -0,0 +1,143 @@
...
+encoding only, followed by a usleep() to maintain the original
+frame-rate timing budget.
```
sinuscosinustan: Misleading. The original `processTime` was a `select()` timeout inside `rfbProcessEvents()` which early returns when there's some IO. `usleep()` does an unconditional sleep. Calling this "maintaining the original timing budget" suggests equivalent behaviour, but it adds fixed latency to every frame cycle even when there's data.

meta-canopy/meta-hpe/meta-common/recipes-graphics/obmc-ikvm/obmc-ikvm/0001-ikvm-decouple-input-thread-from-video-frame-rate.patch
```diff
@@ -0,0 +1,143 @@
...
+ 
+     if (server->clientHead)
+     {
+@@ -92,6 +106,11 @@ void Server::run()
+     }
+ }
+ 
++void Server::pollInput(long int timeout)
++{
++    rfbCheckFds(server, timeout);
```
sinuscosinustan: Please don't do this! `rfcCheckFds` does more than polling the input. It [mutates the client list](https://github.com/LibVNC/libvncserver/blob/009008e2f4d5a54dd71f422070df3af7b3dbc931/src/libvncserver/sockets.c#L359-L362), calls `rfbProcessClientMessages` when there clients with buffered data, and also handles `SetPixelFormat` and such. None of this is thread-safe.

`rfbProcessEvents()` already handles `rfbCheckFds()` and `rfbUpdateClient()`, but it's not being used after this patch anymore. If the goal is to poll more frequently, call `rfbProcessEvents()` from a single thread with a short timeout and protect `sendFrame()` / `doResize()` with a mutex.

Also, if you run `doResize()` from the main thread, it may will create a new framebuffer which is then being re-allocated. If `pollInput` runs while `doResize` happens, it'll cause memory corruption.

Yes. LibVNC is a nightmare...

sinuscosinustan (changes requested):
This version will cause way more issues than the original one and still has the same issue. `libvncserver` is anything but thread-safe, and you're now calling `rfcCheckFds` without any mutexes, which will ultimately cause unexpected issues due to the concurrent access from `rfbUpdateClient`, `sendFrame` and such. The lock-step between server and video threads is also still in place, so the input handling isn't decoupled, but only unsynchronised now.

Disclosure:  Some parts of the review were made with the help of an LLM due to the unreadable `libvncserver` code that is also missing some documentation.

walterchris:
_way more_ is hard :broken_heart: 

Let me dive into this again.

The goal is not to poll more often in that sense - I like to be able to poll the keyboard inputs all the time and decouple this completely from the KVM image.

mynetz:
Is this dead? Or low prio?

mynetz:
@claude review

(lets see if the clanker is listening in)

walterchris:
This is stale for now.

PR #118, opened 2026-03-31
meta-hpe: entity-manager: add registration loop fix

sinuscosinustan:
Decorator objects may be provided by devicetree-vpd-parser (via MachineContext) and used as probe, but then also get populated to the same decorator object, causing a circular loop in registration of objects.

Skip probing interfaces under "xyz.openbmc_project.EntityManager" to prevent circular probe/registration loops.

The patch has been submitted to upstream for review.

### DIFF
--- a/meta-canopy/meta-hpe/meta-common/recipes-phosphor/configuration/entity-manager_git.bbappend
+++ b/meta-canopy/meta-hpe/meta-common/recipes-phosphor/configuration/entity-manager_git.bbappend
@@ -2,6 +2,8 @@ FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
 
 SRC_URI += "file://0001-devicetree-vpd-add-part-number-parsing.patch"
 SRC_URI += "file://0002-devicetree-vpd-add-manufacturer-parsing.patch"
+SRC_URI += "file://0003-entity-manager-move-D-Bus-name-and-path-to-shared-co.patch"
+SRC_URI += "file://0004-entity-manager-avoid-probing-own-D-Bus-name.patch"
 ERROR_QA:remove = "patch-status"
 
 # Enable devicetree VPD parser for platform identification
--- a/meta-canopy/meta-hpe/meta-common/recipes-phosphor/configuration/entity-manager/0003-entity-manager-move-D-Bus-name-and-path-to-shared-co.patch
+++ b/meta-canopy/meta-hpe/meta-common/recipes-phosphor/configuration/entity-manager/0003-entity-manager-move-D-Bus-name-and-path-to-shared-co.patch
@@ -1,0 +1,69 @@
+From 5e61133f12fe2732e243916993892e4314077f3e Mon Sep 17 00:00:00 2001
+From: Tan Siewert <tan.siewert@9elements.com>
+Date: Mon, 13 Apr 2026 14:06:59 +0200
+Subject: [PATCH] entity-manager: move D-Bus name and path to shared constants
+
+entity-manager's D-Bus name and path may be used across multiple files.
+To prevent accidential typos and to increase the consistency, move those
+strings into shared constants.
+
+Upstream-Status: Pending
+Change-Id: I8e455121efb901dd4b70934cc9889772b605b819
+Signed-off-by: Tan Siewert <tan.siewert@9elements.com>
+---
+ src/entity_manager/entity_manager.cpp | 3 +--
+ src/entity_manager/main.cpp           | 3 ++-
+ src/entity_manager/utils.hpp          | 2 ++
+ 3 files changed, 5 insertions(+), 3 deletions(-)
+
+diff --git a/src/entity_manager/entity_manager.cpp b/src/entity_manager/entity_manager.cpp
+index bc78ad0..b6af3ca 100644
+--- a/src/entity_manager/entity_manager.cpp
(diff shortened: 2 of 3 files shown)


meta-canopy/meta-hpe/meta-common/recipes-phosphor/configuration/entity-manager_git.bbappend
```diff
@@ -2,6 +2,7 @@ FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
 
 SRC_URI += "file://0001-devicetree-vpd-add-part-number-parsing.patch"
 SRC_URI += "file://0002-devicetree-vpd-add-manufacturer-parsing.patch"
+SRC_URI += "file://0001-entity-manager-avoid-probing-own-D-Bus-interfaces.patch"
```
walterchris: 0003?

sinuscosinustan:
Marked as draft to prevent merging. There's a bug in my current implementation that causes `propertiesChangedInProgress` to be stuck in `true`, which results in no new scans due to an early return (Thanks Alexander for pointing this possible issue out on Gerrit!)

Edit: Turns out my debug logging was not enough :D It's being set to false properly.

meta-canopy/meta-hpe/meta-common/recipes-phosphor/configuration/entity-manager/0003-entity-manager-avoid-probing-own-D-Bus-name.patch
```diff
@@ -0,0 +1,189 @@
...
++++ b/src/entity_manager/perform_scan.cpp
+@@ -94,16 +94,22 @@ static void processDbusObjects(
+ 
+         for (const auto& [busname, ifaces] : object)
+         {
+-            for (const std::string& iface : ifaces)
++            // We should skip ourselve for probing to avoid circular
++            // probes / registrations when a configuration probe uses
++            // an interface that's also being populated by EM itself.
++            if (busname != emDbusName)
```
walterchris: Wouldn't it be cleaner to do 

```
if (busname == emDBusName)
{
    continue;
}
```

walterchris: NIT: ourselve**s**

meta-canopy/meta-hpe/meta-common/recipes-phosphor/configuration/entity-manager/0003-entity-manager-avoid-probing-own-D-Bus-name.patch
```diff
@@ -0,0 +1,189 @@
...
+entity-manager[257]: Unable to initialize dbus interface : sd_bus_add_object_vtable: org.freedesktop.DBus.Error.FileExists: File exists object Path : /xyz/openbmc_project/inventory/system/chassis/chassis interface name : xyz.openbmc_project.Inventory.Decorator.AssetTag
+entity-manager[257]: Unable to initialize dbus interface : sd_bus_add_object_vtable: org.freedesktop.DBus.Error.FileExists: File exists object Path : /xyz/openbmc_project/inventory/system/chassis/chassis interface name : xyz.openbmc_project.Inventory.Item
+entity-manager[257]: Unable to initialize dbus interface : sd_bus_add_object_vtable: org.freedesktop.DBus.Error.FileExists: File exists object Path : /xyz/openbmc_project/inventory/system/chassis/chassis interface name : xyz.openbmc_project.Inventory.Item.Board.Motherboard
+```
+
+Skip probing our own bus under `xyz.openbmc_project.EntityManager` to
+prevent circular probe/registration loops.
+
+Also refactors hardcoded D-Bus name and path strings into shared
```
walterchris: I would put this into its own commit.

walterchris: (to not creep the scope)

meta-canopy/meta-hpe/meta-common/recipes-phosphor/configuration/entity-manager/0003-entity-manager-avoid-probing-own-D-Bus-name.patch
sinuscosinustan: fair enough. would also allow adding a log info that we skipped ourselves. will fix.

meta-canopy/meta-hpe/meta-common/recipes-phosphor/configuration/entity-manager/0003-entity-manager-avoid-probing-own-D-Bus-name.patch
sinuscosinustan: that was a "discussion" in upstream too ([Gerrit](https://gerrit.openbmc.org/c/openbmc/entity-manager/+/88817/comment/ef1f5db3_157e9398/)). If that's preferred I'll split it up

walterchris: Might be easier to get in then. :)

walterchris (approved):

PR #124, opened 2026-04-06
fix: Serial TX console on HPE Gen11

### LINKED ISSUE #126: HPE ProLiant Gen11: Serial Console Input does not work (SOL)
# Issue Summary

The Input of the SOL does not work - the SOL does not accept any input characters.

## Expected behavior

SOL RX and TX works ;)

## Current behavior

Only SOL RX works

## Workaround

None

walterchris:
Resolves #126


sylv-io (approved):
lgtm!

`Note for transparency:` I requested a change to the kernel patch via chat so that it would use `UART_BUG_TXEN` to resolve the issue.

PR #125, opened 2026-04-06
Fix Build Instructions in README

### LINKED ISSUE #121: README file incorrectly suggests env variable to supply private key
# Issue Summary

In the [Signing key paragraph](https://github.com/canopybmc/canopybmc/blob/main/README.md#signing-key) of the README.md file, the environment variable (export or inline) method to supply the signing key path does not work. Only the `local.conf` method works.

## Suggestions to fix the problem
 
 -  Remove the export and inline methods from the README file.
 - add `HPE_SIGNING_KEY` to `BB_ENV_PASSTHROUGH_ADDITIONS` (Tan suggestion) and test if it works.

walterchris:
Resloves #121

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,19 +58,27 @@ In case this is not provided the image will be signed by the default OSFCI key.
 
 You have several options for this (similar to other variables)
 
-1. Export
+1. `local.conf`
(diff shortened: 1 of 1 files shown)


mynetz (approved):
Thanks for reworking this! Looks good to me.

PR #130, opened 2026-04-07
Canopy: Add git-based version detection

sinuscosinustan:
Replace the hard-coded CANOPY_VERSION variable with a dynamic approach to generate the os-release, inspired by Ampere's approach, that auto-detects the version from the git history and tag at build time.

The version string can still be hard-coded by setting the CANOPY_VERSION variable, though it is not recommended.

The new default will generate the version on the exact tag on HEAD (e.g. `2026.06`, `2026.06-rc1`), but untagged commits fall back to `<nearest-tag>-<commits-ahead>-<shorthash>` via `git describe`.

A custom suffix can be added using the CANOPY_CUSTOM_TAG variable.


walterchris:
It makes sense that the integration test fails. You eithe rhave to sync it manually via the tooling (`fwci sync -f`) or merge it with failing integration test. FirmwareCI does not support branches yet.

walterchris (approved):

PR #135, opened 2026-04-12
Hpe/feature/platdef extract

### LINKED ISSUE #133: Implement PlatDef v1/v2
We need to implement PlatDefv1 commands via CHIF.

Looking into this actually revealed that the problem is a bit deeper than we thought. openBMC does not receive the PlatDef Array through the BIOS (even though there are specific PlatDef Up-/Download Commands) but rather we have to read the PlatDef file from a specific offset from the BIOS flash directly. Once received from the BIOS flash this enables us to:
- Respond to a download request from the BIOS
- Set up i2c-proxy

As a first step we need to:
- Extend the current kernel driver that detects the mtd device to also detect the BIOS flash
- Extend the chif-service to read the mtd device on startup to receive the PlatDev Array and save it for later use
- Implement the PlatDefv1 commands.

walterchris:
Implement the CHIF I2C proxy pipeline: extract the Platform Definition blob from the host BIOS SPI flash, parse I2C segment mappings, map them with the kernel's i2cmux topology, and proxy I2C transactions from BIOS to the hardware during POST.

The individual commits are:
- bf5e469 - scan the host-prime MTD partition for the APML firmware volume, locate the PlatDef FFS file by GUID, and zlib-decompress the record bundle
- 7a418bc - walk PlatDef records (stride = hdr.size × 16), extract I2CEngine segments with CPLD mux register and channel values
- a3633d6 - combine PlatDef segment→CPLD mappings with sysfs i2cmux topology to produce segment→kernel bus number lookups
- 536dbd1 - handle CHIF command 0x0072 by resolving the BIOS segment to a kernel I2C bus and executing I2C_RDWR ioctl transactions
- 0cb3e9a - split the generic UEFI FV/FFS extraction logic into its own compilation unit for clarity and reuse
- 15868eb - Correctly respond to PlatDefv1 download requests with PLATDEF_SMIF_RC_NO_MORE_RECORDS

Tested on the DL365 and DL320.

### DIFF
--- a/meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/platdef_extract.cpp
+++ b/meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/platdef_extract.cpp
@@ -117,6 +117,52 @@ struct PlatDefTableData
 
 static constexpr uint16_t tableDataFlagZLib = 0x0010;
 
+// I2C mux control — CPLD variant (muxType == 1)
+struct PlatDefI2CMuxCpld
+{
+    uint8_t muxType;
+    uint8_t byte;        // CPLD register offset
+    uint8_t initMask;
+    uint8_t selectMask;
+};
+
+// I2C mux control union (16 bytes, only CPLD variant used here)
+union PlatDefI2CMux
+{
+    uint8_t rawBytes[16];
+    uint8_t muxType;
+    PlatDefI2CMuxCpld cpld;
+};
+
+struct PlatDefI2CSegment
+{
+    uint32_t flags;
+    uint8_t speed;
+    uint8_t id;          // BIOS segment byte
+    uint8_t parentId;
+    uint8_t gpuRiserNumber;
+    uint8_t reserved[8];
+    PlatDefI2CMux muxControl;
+};
+
+// PlatDefPrimitive is a 16-byte union in HPE's platdef.h
+struct PlatDefPrimitive
+{
+    uint8_t rawBytes[16];
+};
+
+// I2CEngine fixed fields between the record header and the segments array
+struct PlatDefI2CEngineFixed
+{
+    PlatDefPrimitive reset;
+    PlatDefPrimitive alert;
+    uint32_t flags;
+    uint8_t speed;
+    uint8_t id;
+    uint8_t count;
+    uint8_t reserved[9];
+};
+
 #pragma pack(pop)
 
 static_assert(sizeof(PlatDefBundleHeader) == 32,
@@ -125,6 +171,12 @@ static_assert(sizeof(PlatDefRecordHeader) == 32,
               "PlatDefRecordHeader must be 32 bytes (Gen11 layout)");
 static_assert(sizeof(PlatDefTableData) == 112,
               "PlatDefTableData must be 112 bytes");
+static_assert(sizeof(PlatDefI2CSegment) == 32,
+              "PlatDefI2CSegment must be 32 bytes");
+static_assert(sizeof(PlatDefI2CMux) == 16,
+              "PlatDefI2CMux must be 16 bytes");
+static_assert(sizeof(PlatDefI2CEngineFixed) == 48,
+              "PlatDefI2CEngineFixed must be 48 bytes");
 
 static constexpr uint32_t efiFvhSignature = 0x4856465F; // '_FVH'
 static constexpr uint8_t efiFvFileTypeFreeform = 0x02;
@@ -477,6 +529,115 @@ std::vector<uint8_t> extractPlatDef()
 
     lg2::info("PlatDef: APML firmware volume not found in ROM");
     return {};
+}
+
+// ---------------------------------------------------------------------------
+// PlatDef record parser — extract I2C segment→CPLD mappings
+//
+// Records are walked using hdr.size * 16 as the stride (each record's
+// size field is in units of 16 bytes). We only parse RecordType_I2CEngine
+// (type 14) records; all others are skipped.
+// ---------------------------------------------------------------------------
+
+static constexpr uint8_t recordTypeI2CEngine = 14;
+static constexpr uint8_t recordTypeEndOfTable = 255;
+static constexpr uint8_t i2cMuxTypeCpld = 1;
+
+std::unordered_map<uint8_t, I2cSegmentMapping>
+    parseI2cSegments(const std::vector<uint8_t>& platDef)
+{
+    std::unordered_map<uint8_t, I2cSegmentMapping> result;
+
+    if (platDef.size() < sizeof(PlatDefRecordHeader))
+    {
+        return result;
+    }
+
+    // Validate Gen11 record layout: the first record is the table data
+    // header whose stride should match sizeof(PlatDefTableData).
+    PlatDefRecordHeader firstHdr{};
+    std::memcpy(&firstHdr, platDef.data(), sizeof(firstHdr));
+    if (static_cast<size_t>(firstHdr.size) * 16 != sizeof(PlatDefTableData))
+    {
+        lg2::error("PlatDef: unexpected table record size {SZ} "
+                   "(expected {EXP}), layout may not be Gen11",
+                   "SZ", static_cast<size_t>(firstHdr.size) * 16,
+                   "EXP", sizeof(PlatDefTableData));
+        return result;
+    }
+
+    size_t pos = 0;
+
+    while (pos + sizeof(PlatDefRecordHeader) <= platDef.size())
+    {
+        PlatDefRecordHeader hdr{};
+        std::memcpy(&hdr, platDef.data() + pos, sizeof(hdr));
+
+        if (hdr.type == recordTypeEndOfTable || hdr.size == 0)
+        {
+            break;
+        }
+
+        size_t recordBytes = static_cast<size_t>(hdr.size) * 16;
+
+        if (pos + recordBytes > platDef.size())
+        {
+            lg2::warning("PlatDef: record at 0x{OFF} claims {SZ} bytes, "
+                         "only {REM} remain",
+                         "OFF", lg2::hex, pos, "SZ", recordBytes,
+                         "REM", platDef.size() - pos);
+            break;
+        }
+
+        if (hdr.type == recordTypeI2CEngine)
+        {
+            size_t fixedEnd = pos + sizeof(PlatDefRecordHeader) +
+                              sizeof(PlatDefI2CEngineFixed);
+            if (fixedEnd > platDef.size())
+            {
+                break;
+            }
+
+            PlatDefI2CEngineFixed eng{};
+            std::memcpy(&eng, platDef.data() + pos +
+                                  sizeof(PlatDefRecordHeader),
+                        sizeof(eng));
+
+            size_t recordEnd = pos + recordBytes;
+            size_t segStart = fixedEnd;
+
+            for (uint8_t i = 0; i < eng.count; i++)
+            {
+                size_t segOff = segStart + i * sizeof(PlatDefI2CSegment);
+                if (segOff + sizeof(PlatDefI2CSegment) > platDef.size() ||
+                    segOff + sizeof(PlatDefI2CSegment) > recordEnd)
+                {
+                    break;
+                }
+
+                PlatDefI2CSegment seg{};
+                std::memcpy(&seg, platDef.data() + segOff, sizeof(seg));
+
+                if (seg.muxControl.muxType == i2cMuxTypeCpld)
+                {
+                    result[seg.id] = {seg.muxControl.cpld.byte,
+                                      seg.muxControl.cpld.selectMask};
+
+                    lg2::debug("PlatDef I2C: engine={ENG} seg=0x{SEG} "
+                               "cpld=0x{CPLD} chan={CH}",
+                               "ENG", eng.id, "SEG", lg2::hex, seg.id,
+                               "CPLD", lg2::hex, seg.muxControl.cpld.byte,
+                               "CH", seg.muxControl.cpld.selectMask);
+                }
+            }
+        }
+
+        pos += recordBytes;
+    }
+
+    lg2::info("PlatDef: parsed {CNT} I2C segment mappings",
+              "CNT", result.size());
+    return result;
 }
 
 } // namespace chif
--- a/meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/platdef_extract.cpp
+++ b/meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/platdef_extract.cpp
@@ -1,12 +1,14 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright (C) 2026 9elements GmbH
 
-// Extract the APML Platform Definition blob from the host BIOS SPI flash.
-// The blob is stored inside a UEFI firmware volume as an FFS file,
-// identified by well-known GUIDs. This replaces HPE's standalone
-// hpe-platdef-extract utility with an in-process implementation.
+// Parse the HPE Platform Definition (PlatDef) blob and extract I2C
+// segment-to-bus mappings for the CHIF I2C proxy. The raw blob is
+// extracted from the host BIOS SPI flash via the UEFI FV parser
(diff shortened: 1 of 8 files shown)


meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/platdef_extract.cpp
```diff
@@ -0,0 +1,570 @@
...
+
+    // Compressed: zlib data starts after the PlatDefTableData header
+    size_t compOffset = offset + sizeof(PlatDefTableData);
+    if (tableHdr.compressedSize < sizeof(PlatDefTableData))
+    {
+        lg2::error("PlatDef: invalid compressedSize {SZ}",
+                   "SZ", tableHdr.compressedSize);
+        return {};
+    }
+    uLong compSize = tableHdr.compressedSize - sizeof(PlatDefTableData);
```
sinuscosinustan: the mixed use of `uLong` and `uint32_t` is a bit meh. Please make it consistent to `uint32_t` (I am wondering where `uLong` is coming from anyway)

meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/platdef_extract.cpp
```diff
@@ -0,0 +1,570 @@
...
+
+    if (compOffset + compSize > raw.size())
+    {
+        lg2::error("PlatDef: compressed data exceeds blob size");
+        return {};
+    }
+
+    // Decompress into output buffer: [PlatDefTableData][decompressed records]
+    uLong decompSize = tableHdr.totalSize;
```
sinuscosinustan: same here

meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/platdef_extract.cpp
```diff
@@ -0,0 +1,570 @@
...
+#pragma pack(push, 1)
+
+struct PlatDefBundleHeader
+{
+    char signature[16]; // "$PlatdefBundle$1"
+    uint32_t totalSize;
+    uint16_t flags;
+    uint8_t headerLength;
+    uint8_t count;
+    uint8_t reserved[8];
```
sinuscosinustan, line 39: tho reserved, it should be an `uint32_t` if we want to match with the original implementation (and because we have the pack pragma set, which could break parsing if the header actually contains all 32 bytes instead of the 26)

meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/platdef_extract.cpp
```diff
@@ -0,0 +1,570 @@
...
+    uint8_t reserved[10];
+};
+
+static constexpr uint16_t tableDataFlagZLib = 0x0010;
+
+// I2C mux control — CPLD variant (muxType == 1)
+struct PlatDefI2CMuxCpld
+{
+    uint8_t muxType;
+    uint16_t byte;       // CPLD register offset
```
sinuscosinustan: that should be a `uint8_t`

meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/smif_service.cpp
```diff
@@ -296,6 +307,185 @@ int SmifService::handleGetEvAuthStatus(const ChifPktHeader& hdr,
...
+
+    struct i2c_msg msgs[2];
+    int numMsgs = 0;
+    uint8_t writeBuf[32] = {};
+    uint8_t readBuf[32] = {};
+
+    if (writeLen > 0)
+    {
+        size_t available = reqPayload.size() - i2cOffData;
+        size_t copyLen = std::min(static_cast<size_t>(writeLen), available);
```
sinuscosinustan: could end up in out-of-bounds. Please return if `writeLen > available`.

meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/smif_service.cpp
```diff
@@ -296,6 +307,185 @@ int SmifService::handleGetEvAuthStatus(const ChifPktHeader& hdr,
...
+    int busNum = it->second;
+    uint8_t i2cAddr = static_cast<uint8_t>(address >> 1);
+
+    char devPath[32];
+    std::snprintf(devPath, sizeof(devPath), "/dev/i2c-%d", busNum);
+
+    int fd = open(devPath, O_RDWR);
+    if (fd < 0)
+    {
+        uint32_t err = i2cGeneralError;
```
sinuscosinustan, line 423: can you add a warning or so if this failed? (In general, the whole `handleI2cProxy` method does not have much logging)

meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/smif_service.cpp
```diff
@@ -296,6 +307,185 @@ int SmifService::handleGetEvAuthStatus(const ChifPktHeader& hdr,
...
+        uint32_t err = i2cSegmentNotFound;
+        std::memcpy(resp.data() + i2cRespOffError, &err, sizeof(err));
+        return i2cRespSize;
+    }
+
+    int busNum = it->second;
+    uint8_t i2cAddr = static_cast<uint8_t>(address >> 1);
+
+    char devPath[32];
+    std::snprintf(devPath, sizeof(devPath), "/dev/i2c-%d", busNum);
```
sinuscosinustan: I am not a huge fan of such static allocs but this might be a personal preference. would see it as a nit

meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/platdef_extract.cpp
```diff
@@ -0,0 +1,570 @@
...
+
+                PlatDefI2CSegment seg{};
+                std::memcpy(&seg, platDef.data() + segOff, sizeof(seg));
+
+                if (seg.muxControl.muxType == i2cMuxTypeCpld)
+                {
+                    auto cpldReg = static_cast<uint8_t>(
+                        seg.muxControl.cpld.byte);
+                    result[seg.id] = {cpldReg,
+                                      seg.muxControl.cpld.initMask};
```
sinuscosinustan: shouldn't this be `selectMask`? (at least I think if I looked at the right part of the original implementation)

meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/platdef_extract.cpp
```diff
@@ -0,0 +1,570 @@
...
+// sysfs gives us:   (muxIndex, channelNumber) → kernelBus
+// Correlation:      muxIndex = cpldRegister - cpldRegisterBase
+// ---------------------------------------------------------------------------
+
+// GXP AHB bus path where i2cmux devices appear in sysfs
+static constexpr auto ahbSysfsPath = "/sys/devices/platform/ahb@80000000";
+
+// CPLD register base address for mux index 0. Mux N maps to register
+// base + N. Derived from the device tree mux-reg-masks property.
+static constexpr uint8_t cpldRegisterBase = 0x84;
```
sinuscosinustan, line 439: is it really always `0x84` or just coincidence? I can't see any static definition for it in the original impl. (ctx: it was queried from a text file in the original impl in `/tmp`)

meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/smif_service.cpp
```diff
@@ -296,6 +307,185 @@ int SmifService::handleGetEvAuthStatus(const ChifPktHeader& hdr,
...
+
+    return i2cRespSize;
+}
+
+// ---------------------------------------------------------------------------
+// PlatDef v1 download handler (0x0200)
+// ---------------------------------------------------------------------------
+
+static constexpr uint32_t platDefNoMoreRecords = 6;
+static constexpr size_t platDefPayloadSize = 4024;
```
sinuscosinustan, line 497: why's the size hard-coded and not some sort of `sizeof` of the relevant packet struct?

meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/platdef_extract.cpp
```diff
@@ -0,0 +1,570 @@
...
+            return "/dev/" + entry.path().filename().string();
+        }
+    }
+    return {};
+}
+
+// Decompress the PlatDef blob from the raw APML file.
+// Layout: [PlatDefBundleHeader][PlatDefTableData][zlib compressed records...]
+// Returns: [PlatDefTableData header][decompressed records]
+static std::vector<uint8_t> decompressPlatDef(const std::vector<uint8_t>& raw)
```
sinuscosinustan, line 177: Is there any signature available? I can see that the original implementation didn't do such thing, but they still defined `PlatDefBundleSignature` and `PlatDefBundleEndMarker`

meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/uefi_fv.cpp
```diff
@@ -0,0 +1,236 @@
...
+           (static_cast<uint32_t>(hdr.size[1]) << 8) |
+           (static_cast<uint32_t>(hdr.size[2]) << 16);
+}
+
+std::vector<uint8_t> extractFfsFile(std::istream& rom, size_t romSize,
+                                     const EfiGuid& fvGuid,
+                                     const EfiGuid& fileGuid)
+{
+    for (size_t fvOffset = 0; fvOffset + sizeof(EfiFvHeader) < romSize;
+         fvOffset += 0x10000)
```
sinuscosinustan: can we increase it by using the actual length given from the firmware volume header (aligned w/ 64k)? this seems to be done in the original impl too.

meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/smif_service.cpp
```diff
@@ -296,6 +307,185 @@ int SmifService::handleGetEvAuthStatus(const ChifPktHeader& hdr,
...
+static constexpr auto i2cRespSize =
+    static_cast<uint16_t>(sizeof(ChifPktHeader) + sizeof(I2cResponse));
+static constexpr size_t i2cRespOffError = offsetof(I2cResponse, errcode);
+static constexpr size_t i2cRespOffAddress = offsetof(I2cResponse, address);
+static constexpr size_t i2cRespOffSegment = offsetof(I2cResponse, engine);
+static constexpr size_t i2cRespOffReadLen = offsetof(I2cResponse, readLen);
+static constexpr size_t i2cRespOffData = offsetof(I2cResponse, data);
+
+// I2C error codes
+static constexpr uint32_t i2cGeneralError = 7;
```
sinuscosinustan: that should be `101`? see `include/i2c_return_codes.hpp`

sinuscosinustan (changes requested):
This is already way more comfortable to read and to review than the original implementation, so thanks for tackling through this!

A few points, some of which are actual bugs / mismatches from the original impl. I may or may not have missed some parts, so please bear with me if I report something I haven't noticed yet later the day.

meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/platdef_extract.cpp
walterchris: uLong is from zlib - but yeah.

meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/smif_service.cpp
walterchris: Looking at the `0001-apply-HPE-custom-changes.patch` it says:

```
+/* 0x8072 - I2C Transaction Response
+ * errorcode values:
+ * 0: success
+ * 1: transaction timeout
+ * 2: invalid engine
+ * 3: invalid target
+ * 4: invalid read count
+ * 5: invalid write count
+ * 6: Bad magic signature
+ * 7: generic I2C error
```

However cross-checking now with https://github.com/HewlettPackard/openbmc-chif-svc/blob/75f98cc221bd17a9ed00cd8b404ccccd2183934a/src/smif.cpp#L937 it's most likely: `7:    Unauthorized. Must be in factory mode.`

Let me change it to `101` - even though that is not mentioned there at all.

meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/smif_service.cpp
walterchris: IF we don't send the full 4024 bytes the BIOS errors. I tried sending only the necessary length but it looks like we always need to send the full 4024 bytes.

walterchris: So it's always: `[errorCode u32][recordCount u32][recordData 4016].` - I added a comment.

meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/platdef_extract.cpp
walterchris: Yes - the text file was build up dynamically. THe 0x84 is true for all platforms I tested but we should derive this dynamically. Let me create a follow-up issue for this.

walterchris: https://github.com/canopybmc/canopybmc/issues/138

meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/platdef_extract.cpp
walterchris: Where do you got the `unit32_t` suggestion from?

HPE upstream suggest:

```
typedef struct {
    char        Signature[16];
    UINT32      TotalSize;
    UINT16      Flags;
    UINT8       HeaderLength;   // header size in units of 16 bytes
    UINT8       Count;          // number of PlatformId entries
    UINT16      PlatformId[1];  // variable-length array
} PlatDefBundleHeader;
```

So the base struct would be 26 bytes - HeaderLength is normally 2 * 16 so 32 bytes.

If we really want to improve it we would need to parse `HeaderLength` correctly. Do you want me to do this here? Or follow-up in another PR on this?

walterchris:
I (force) pushed the commits with the changes.

meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/platdef_extract.cpp
sinuscosinustan: You're right! The original code is giving me a headache, ngl. Originally I wanted to write `uint16_t` instead of `uint32_t` (that was a brainglitch from me), but I completely ignored the fact that it's a variable-length array.

sinuscosinustan (approved):

PR #139, opened 2026-04-15
HPE: re-arm boot gate after pgood de-assertion

### LINKED ISSUE #136: HPE Proliant Gen11: Investigate Power Control
# Issue Summary

We need to investigate if all power control commands (via obmcutil and Redfish) work properly. I just noticed that GracefulReboot results in the host being powered off. There might be more here as we did not check the power states yet in great detail.

## Expected behavior

Power Control works correctly.

## Current behavior

I just noticed that when triggering GracefulReboot via Redfish the host stays off.

## Workaround

No.

sinuscosinustan:
When PGOOD de-asserts, `gxp-power-ctrl` correctly held the boot gate and cycled `soc_release`, but it never re-armed the boot gate. This prevented the CPLD from bringing PGOOD back after a UEFI-initiated power cycle (triggered by `HpPowerCycleDxe`), or a graceful restart initiated by the user via the BMC, so the IRQ never fired and `x86-power-control` could never transition back to the "On" state.

Call `gxp_power_ctrl_prepare_boot()` immediately after `gxp_power_ctrl_shutdown_ack()` in the PGOOD IRQ handler, matching the behaviour of the original HPE userspace-based script implemenation which unconditionally locks the gate and then re-arms it before waiting for PGOOD to return.


walterchris (approved):

walterchris:
There is a type in the commit and patch file - it should be implementation.

walterchris (changes requested):
Fix typo :)

sinuscosinustan:
@walterchris fixed. also fixed the usual mistake of mine by using the British version of "behaviour" in a commit message

walterchris (approved):

meta-canopy/meta-hpe/meta-common/recipes-x86/chassis/x86-power-control/power-config-host0.json
```diff
@@ -37,6 +37,7 @@
       "ResetPulseMs": 500,
       "PowerCycleMs": 5000,
       "SioPowerGoodWatchdogMs": 1000,
+      "PowerOkWatchdogMs": 8000,
```
sylv-io: `PowerOKWatchdogMs` (`Ok` -> `OK`)

meta-canopy/meta-hpe/meta-gxp/recipes-kernel/linux/linux-stable-6.18/0082-soc-hpe-gxp-power-ctrl-re-arm-boot-gate-after-PGOOD-.patch
```diff
@@ -0,0 +1,43 @@
...
+diff --git a/drivers/soc/hpe/gxp-power-ctrl.c b/drivers/soc/hpe/gxp-power-ctrl.c
+index 91e07803f8d4..8f40bc3200ce 100644
+--- a/drivers/soc/hpe/gxp-power-ctrl.c
++++ b/drivers/soc/hpe/gxp-power-ctrl.c
+@@ -316,8 +316,10 @@ static irqreturn_t gxp_power_ctrl_pgood_irq(int irq, void *data)
+ 	if (pgood <= 0) {
+ 		/* PGOOD fell: host powered off, hold boot gate */
+ 		gxp_power_ctrl_shutdown_ack(pctrl);
++		/* Re-arm boot gate */
++		gxp_power_ctrl_prepare_boot(pctrl);
```
sylv-io: i am not sure if we should reuse `gxp_power_ctrl_prepare_boot()` here, since it also resets `CSM_SHUTDOWN_REASON` 

should we define and use `gxp_power_ctrl_rearm_boot_gate()` as an alternative instead?

sylv-io (changes requested):

meta-canopy/meta-hpe/meta-common/recipes-x86/chassis/x86-power-control/power-config-host0.json
sinuscosinustan: Fixed locally but never pushed :woman_facepalming:

meta-canopy/meta-hpe/meta-gxp/recipes-kernel/linux/linux-stable-6.18/0082-soc-hpe-gxp-power-ctrl-re-arm-boot-gate-after-PGOOD-.patch
sinuscosinustan: Fair point, will do that.

sylv-io (approved):
lgtm

PR #143, opened 2026-04-15
gxp-udc: fix HID gadget failure after BIOS-to-OS EHCI handoff

### LINKED ISSUE #142: GXP UDC can not be enumerated by the OS after BIOS hand-off
# Issue Summary

The GXP SoC has a virtual USB Hub Controller that sits between the host's EHCI ports and the GXP.  When the BIOS hands off to the OS, the OS resets the EHCI controller in order to be able to enumerate it. However the VUHC absorbs the reset internally and does not propagate it has PORTRST or PORTEN_CHANGE interrupt. The device is left in a stale state - still BIOS assigned address (EVDADDR=1), old endpoint configuration, connect bit still set.

The host fails to enumerate the device and thus Inputs via KVM does not work in the OS.

## Expected behavior

Input should work after BIOS hand off. The kernel code should be able to handle this properly (We see the port_owner bit changes during handoff - however nothing else happens.

## Current behavior

Keyboard inputs don't work after BIOS hand off.

## Workaround

The current work around is to reconnect to KVM. It will trigger a rebind (unbind/bind) and this will enable the Host to enumerate the device properly.

walterchris:
The HID device becomes non-functional after BIOS hands off control to the OS. The issue is that the VUHC does silently drops the reset from the OS and disables the UDC port after ownership transfer. The device is left in a stale state.

This PR does three things:
- 0083: It detects via a watchdog timer if the port died. Unfortunately we can not register any interrupts that could potentially handle this - so we have to poll this.
- 0080: It properly resets all states when rerying and does a full disconnect + connect
- 0079: Ensures that data toggles are correct after reconnecting

I use the patch numbers 0079, 0080 and 0083 to fill the gaps :melting_face:


walterchris:
We could probably upstream this?

meta-canopy/meta-hpe/meta-gxp/recipes-kernel/linux/linux-stable-6.18/0080-usb-gadget-gxp-udc-fix-connect-retry.patch
```diff
@@ -0,0 +1,140 @@
...
+-		GXP_UDC_CONNECT_MAX_RETRIES);
+-
+-	/* Toggle connect bit: disconnect */
+-	evdcs &= ~EVDCS_CONNECT;
+-	writel(evdcs, drvdata->base + EVDCS);
++		drvdata->vdevnum, drvdata->connect_retries,
++		GXP_UDC_CONNECT_MAX_RETRIES);
++
++	/* Full disconnect -- reset all device state */
++	gxp_udc_disconnect(drvdata);
```
sylv-io, line 91: iiuc `gxp_udc_disconnect(drvdata)` only masks future interrupts. it does not wait for an existing `gxp_udc_irq()` handler to finish.  
We need to wait until any existing interrupt handler has finished before tearing down endpoints.

```
spin_unlock_irqrestore(&drvdata->lock, flags);
synchronize_irq(drvdata->irq);
spin_lock_irqsave(&drvdata->lock, flags);
```

sylv-io (changes requested):
This is probably unrelated to the issue that this patch addresses, but I noticed some pre-existing issues with the way this driver enforces endpoint queue locking.
Perhaps we should address them in a separate PR.

For now, I only addressed the occasion introduced by this PR.

sylv-io:
> This is probably unrelated to the issue that this patch addresses, but I noticed some pre-existing issues with the way this driver enforces endpoint queue locking. Perhaps we should address them in a separate pull request.
> 
> For now, I only addressed the occasion introduced by this PR.

Here is an untested patch showing how we could resolve these inconsistencies. I'm not sure if you want to address them in this PR or if we should do it in a separate one. :monkey: 

[0001-usb-gadget-gxp-udc-fix-endpoint-queue-locking.patch](https://github.com/user-attachments/files/26903561/0001-usb-gadget-gxp-udc-fix-endpoint-queue-locking.patch)

walterchris:
> > This is probably unrelated to the issue that this patch addresses, but I noticed some pre-existing issues with the way this driver enforces endpoint queue locking. Perhaps we should address them in a separate pull request.
> > For now, I only addressed the occasion introduced by this PR.
> 
> Here is an untested patch showing how we could resolve these inconsistencies. I'm not sure if you want to address them in this PR or if we should do it in a separate one. 🐒
> 
> [0001-usb-gadget-gxp-udc-fix-endpoint-queue-locking.patch](https://github.com/user-attachments/files/26903561/0001-usb-gadget-gxp-udc-fix-endpoint-queue-locking.patch)

I would actually move this into another PR.

meta-canopy/meta-hpe/meta-gxp/recipes-kernel/linux/linux-stable-6.18/0080-usb-gadget-gxp-udc-fix-connect-retry.patch
walterchris: Fixed.

sylv-io (approved):

PR #145, opened 2026-04-16
ci: use GXP2loader image directly and capture all build variants

mynetz:
The generic build action now uploads all obmc-phosphor-image-<board>.*static.mtd artifacts, capturing variant images (e.g. GXP2loader) without leaking platform-specific details. The HPE workflow uses the GXP2loader image directly for FirmwareCI, removing the bootblock concatenation step.

- Generic build action captures all obmc-phosphor-image-<board>.*static.mtd variants in a single artifact, excluding date-stamped images                                                                                                   
- HPE workflow uploads GXP2loader image directly to FirmwareCI, removing the bootblock fetch/concatenation steps and the hpe-gxp-bootblock repo dependency                                                                                 
- QEMU workflow updated to match new artifact naming

### DIFF
--- a/.github/actions/build/action.yml
+++ b/.github/actions/build/action.yml
@@ -45,6 +45,6 @@ runs:
     - name: Upload firmware image
       uses: actions/upload-artifact@v4
       with:
-        name: obmc-phosphor-image-${{ inputs.board }}.static.mtd
-        path: build/${{ inputs.board }}/tmp/deploy/images/${{ inputs.board }}/obmc-phosphor-image-${{ inputs.board }}.static.mtd
+        name: obmc-phosphor-image-${{ inputs.board }}
+        path: build/${{ inputs.board }}/tmp/deploy/images/${{ inputs.board }}/obmc-phosphor-image-${{ inputs.board }}.*static.mtd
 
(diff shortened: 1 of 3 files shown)


walterchris (approved):
I'm not super satisfied with the catch-all.

Even though to have the three images makes sense (depending on which hardware you want to use) the image without the GXP actually does not make sense.

However that might be a different issue: Maybe we should just not build the image without GXP anymore to work around this.

mynetz:
> I'm not super satisfied with the catch-all.

Not using the catch-all will force us to rework the generic builder workflow to know platform specifics. I think this approach right now is a good compromise.

> Even though to have the three images makes sense (depending on which hardware you want to use) the image without the GXP actually does not make sense.
> 
> However that might be a different issue: Maybe we should just not build the image without GXP anymore to work around this.

This needs to be a different discussion. Technically the image without the GXP is what we build and then we have a post action to glue the "right" GXP to the image. This was the way to have _one_ build target for all Gen11 systems. Hence this image will always be build and then modified.

We can maybe rename it in the post step to move it out the way.

This generally relates to the discussion on what is "our" image. We also face this with the HPE "return to iLO" payload and how we structured the image. Our "bmc" section is 32 MiB - 512 KiB (gxp boot section). Their image contains the full 32 MiB. Which also means binary magic. Maybe we have to rethink that.

However, the change to the CI is at the end of this discussion. I worked with what the build provides.

PR #147, opened 2026-04-17
HPE Proliant Gen11: Fix EV Storage return codes and add host to dl320 dut config

walterchris:
We need to return proper EV Storage return codes in order to prevent a reboot loop on the DL320. Weirdly enough this does not happen on the DL365.

So this PR changes the error return codes to be similar with the original HPE definition.

I also fixed the FirmwareCI DL320 dut configuration with this PR.

### DIFF
--- a/meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/smif_service.cpp
+++ b/meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/smif_service.cpp
@@ -90,24 +90,29 @@ int SmifService::buildEvDataResponse(const ChifPktHeader& hdr,
 // EV command handlers
 // ---------------------------------------------------------------------------
 
+constexpr uint32_t ev(EvError e)
+{
+    return static_cast<uint32_t>(e);
+}
+
 int SmifService::handleGetEvByIndex(const ChifPktHeader& hdr,
                                     std::span<const uint8_t> reqPayload,
                                     std::span<uint8_t> response)
 {
     if (!evStorage_ || reqPayload.size() < sizeof(uint32_t))
     {
-        return buildSimpleResponse(hdr, response, 1);
+        return buildSimpleResponse(hdr, response, ev(EvError::deviceError));
     }
 
     uint32_t index = 0;
     std::memcpy(&index, reqPayload.data(), sizeof(index));
 
-    auto ev = evStorage_->getByIndex(index);
-    if (!ev)
+    auto entry = evStorage_->getByIndex(index);
+    if (!entry)
     {
-        return buildSimpleResponse(hdr, response, 1);
+        return buildSimpleResponse(hdr, response, ev(EvError::noSuchEv));
     }
-    return buildEvDataResponse(hdr, response, *ev);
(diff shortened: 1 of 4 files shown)


meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/smif_service.cpp
```diff
@@ -105,7 +105,7 @@ int SmifService::handleGetEvByIndex(const ChifPktHeader& hdr,
     auto ev = evStorage_->getByIndex(index);
     if (!ev)
     {
-        return buildSimpleResponse(hdr, response, 1);
+        return buildSimpleResponse(hdr, response, 2); // no such EV
```
sylv-io: can we use something like enums instead of magic val  + comments to describe  de EV codes?

meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/smif_service.cpp
```diff
@@ -242,11 +242,12 @@ int SmifService::handleEvStats(const ChifPktHeader& hdr,
                                           evStorage_->remainingSize())
                                     : 0;
     auto cnt = static_cast<uint16_t>(evStorage_ ? evStorage_->count() : 0);
-    auto maxSzKb = static_cast<uint16_t>(EvStorage::maxSize() / 1024);
+    // HPE sends raw byte count (65536), not divided by 1024
```
sylv-io: Are your sure? how did you verify this claim?

meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/smif_service.cpp
```diff
@@ -242,11 +242,12 @@ int SmifService::handleEvStats(const ChifPktHeader& hdr,
...
+    auto maxSz = static_cast<uint16_t>(EvStorage::maxSize());
```
sylv-io: iiuc:
```
EvStorage::maxSize() = 65536
```
so this cast will result in: `maxSz = 0`

are you sure this fixes the issue?

sylv-io (changes requested):

meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/smif_service.cpp
walterchris: I droped the whole size change completely - I need to investigate more in this. Let keep the scope for now on the error codes as this fixes the issue I am currently seeing.

meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/smif_service.cpp
walterchris: See above.

sylv-io (approved):

PR #148, opened 2026-04-18
HPE Proliant Gen11: add Field Access Command Handler to the CHIF service

walterchris:
The BIOS reads Serial Number and Product ID via the CHIF Field Access Command (0x0153) and display them in the BIOS.

This PR also fixes two small nits:
- FirmwareCI tests: We now accept 0x2f8 and 0x3f8 as valid LPC addresses for the serial; and we wait a bit longer until the SMBIOS data has been parsed
- Fix a logging error in the CHIF service so that we can properly see the segments and addresses.

### DIFF
--- a/meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/smif_service.cpp
+++ b/meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/smif_service.cpp
@@ -537,7 +537,7 @@ int SmifService::handleI2cProxy(const ChifPktHeader& hdr,
     if (rc < 0)
     {
         int savedErrno = errno;
-        lg2::warning("I2C proxy: ioctl failed seg=0x{SEG} addr=0x{ADDR} "
+        lg2::warning("I2C proxy: ioctl failed seg={SEG} addr={ADDR} "
                      "bus={BUS}: {ERR}",
                      "SEG", lg2::hex, segment, "ADDR", lg2::hex, i2cAddr,
                      "BUS", busNum, "ERR", strerror(savedErrno));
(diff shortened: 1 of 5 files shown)


meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/smif_service.cpp
```diff
@@ -14,14 +14,76 @@
...
+    productId_ = deriveShortProductId(
+        readDtString("/proc/device-tree/model"));
+    if (productId_.empty())
+    {
+        productId_ = "UNKNOWN";
+        lg2::warning("No model in device-tree, using fallback");
+    }
+
+    lg2::info("Platform serial: {SN}, product ID: {PID}", "SN",
+              boardSerial_, "PID", productId_);
```
sinuscosinustan, line 85: wouldn't it make sense to get those data from EM instead or should the CHIF service stay independent of it?

May also be a question for upstream since we already get those data via `devicetree-vpd-parser` (which are then consumed by EM)

walterchris: True - We _could_ do that. I don't have a strong preference over one or the other as both use the same source and the information is static. I slight con on this would be that one could argue that you don't have any startup dependencies. But I don't have a strong preference.

walterchris: Any preference on how to proceed here within this change?

sinuscosinustan: I'd be fine to do it this way for now. We can check on that later if preferred.

sinuscosinustan (approved):

PR #149, opened 2026-04-20
HPE Proliant Gen11: add Security State message handling to CHIF service

### LINKED ISSUE #150: HPE Proliant Gen11: Report back Security State as production
HPE's openBMC hardcodes the Security State as production. Do the same in our implementation. Also log the report state from the BIOS in the logs.

walterchris:
This commit adds the Security State handling to the existing chif service. It _always_ replies to the security state with production, and logs the value when the BIOS tries to set it.

### DIFF
--- a/meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/smif_service.cpp
+++ b/meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/smif_service.cpp
@@ -586,6 +586,50 @@ int SmifService::handlePlatDefDownload(const ChifPktHeader& hdr,
 }
 
 // ---------------------------------------------------------------------------
+// Security state handlers (0x0139 get, 0x0158 set)
+// ---------------------------------------------------------------------------
+
+int SmifService::handleSecurityStateGet(const ChifPktHeader& hdr,
+                                        std::span<uint8_t> response)
+{
+    // pkt_8139: rcode(4) + security_state(1) + reserved(3) = 8 bytes payload
+    constexpr size_t payloadSize = 8;
+    constexpr auto respSize =
+        static_cast<uint16_t>(sizeof(ChifPktHeader) + payloadSize);
+
+    if (response.size() < respSize)
+    {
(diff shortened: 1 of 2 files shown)


meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/smif_service.cpp
```diff
@@ -523,6 +523,50 @@ int SmifService::handlePlatDefDownload(const ChifPktHeader& hdr,
...
+        return -1;
+    }
+
+    std::fill_n(response.data(), respSize, uint8_t{0});
+    initResponse(response, hdr, respSize);
+
+    auto resp = responsePayload(response);
+    // ErrorCode = 0 (already zeroed)
+    resp[sizeof(uint32_t)] =
+        static_cast<uint8_t>(SecurityState::production);
```
sinuscosinustan, line 611: So if the BIOS sends a different security state (e.g. `highSecurity`), we'll still reply with `production`?

walterchris: We are at least mimicking HPEs behavior here - yes.

The alternative would be that we store the value. Either in RAM or persistent.

sinuscosinustan: If it's the actual state of the HPE tarball, then I'd say open a follow-up issue and check back with HPE if it's intentional by them?

sinuscosinustan (approved):
code-wise LGTM. Regarding the security state response I'd suggest to check with HPE

meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/smif_service.cpp
walterchris: Following up with https://github.com/canopybmc/canopybmc/issues/153

Issue #112, opened 2026-03-25
HPE Proliant Gen11: Observations while manual long-time testing `2026.04-rc1`


mynetz:
# Issue Summary

These observations might be issues or not. We collected them while at Cloudfest. Need to be discussed after Cloudfest and transferred into individual issues.

## Issues list

- [x] Inventory LED/Button pressed from chassis has no event in the BMC/web
    - Partially solved (physical UID button still not working). See: canopybmc/canopybmc#113 
- [x] ~~BMC Dump contains a log with wrong encoding (shows only Chinese characters)~~
    - The issue occurs with GNOME text editor. Checking the file with `less` shows that there are some control characters printed by the UEFI, but the main content can be displayed.
- [x] Fan spinup due to CPU/DRAM temp reported to high. Probably PECI problem
    - See: canopybmc/canopybmc#152
    - webui-vue issue - PSM reports the state as off but the UI does not automatically refresh.
- [x] KVM keyboard usage (several observations)
    - using kvm keyboard is better than `hpe-openbmc` but still not great UX.
    - keyboard sometimes register keypress multiple times
    - kvm sometime stop tracking input until a browser reload
    - KVM keyboard sometimes fails to register F-key inputs
    - USB device for keyboard fails to register under unknown circumstances (no inputs at all until BMC is being rebooted)
    - See: canopybmc/canopybmc#114
- [x] Overvoltage error threshold maybe to low (240V) 
    - Explained: This is the max range printed on the supply. More might be acceptable in the error margin.
- [x] EM: circular loop in registration of objects
    - See canopybmc/canopybmc#118 

## Environment

- HPE ProLiant Gen11 DL320 (Nr. 2)
- Manual build from Marcello and Tan

mynetz:
Seems to be all done.

PR #152, opened 2026-04-20
meta-hpe: linux-gxp: prevent overlapping PECI transfers

sylv-io:
This PR addresses the temperature reporting issue related to PECI, which is tracked in issue #112.
The issue observed in PR #112 is unexpected fan spin-up, caused by CPU/DRAM temperatures being reported as too high during long-term manual testing.

While PR #116 adds filtering for suspicious readings, this PR provides a more permanent solution by addressing the underlying issue with the PECI controller rather than merely filtering the symptom.

Further extended testing is required to confirm that this change resolves the incorrect PECI temperature readings and related fan spin-up behaviour observed in issue #112.


sylv-io:
@walterchris:  I tested both 2026.04-rc2 and this PR on the DL320 at https://osfci.tech/ci/. I can confirm that I no longer see any bogus sensor readings.

I entered this command into the BMC serial console while powering on the host to produce the false readings:
```
sh -c '
while :; do
  ts=$(date +%s)
  for d in /sys/class/hwmon/hwmon*; do
    n=$(cat "$d/name" 2>/dev/null) || continue
    case "$n" in
      peci_cputemp.cpu*|peci_dimmtemp.cpu*)
        for f in "$d"/temp*_input; do
          [ -r "$f" ] || continue
          cat "$f" >/dev/null 2>&1 & p=$!
          kill -TERM "$p" 2>/dev/null
          wait "$p" 2>/dev/null || true
          v=$(cat "$f" 2>/dev/null) || continue
          [ -n "$v" ] || continue
          [ "$v" -eq "$v" ] 2>/dev/null || continue
          if [ "$v" -gt 110000 ] || [ "$v" -lt -10000 ]; then
            printf "%s OUTLIER %s=%s\n" "$ts" "$f" "$v"
          fi
        done
        ;;
    esac
  done
done
'
```

After ~30 min of testing, I could only detect outlier sensor readings in the command output or event logs when running 2026.04-rc2 but not after this fix.

walterchris:
Please rebase - then I can approve and merge it.

walterchris (approved):

PR #155, opened 2026-04-20
fix: Increase timeouts for CHIF related tests

walterchris:


sinuscosinustan (approved):

walterchris:
Need to rework this a bit more.

We are now polling until the SMBIOS tables arrive instead of waiting a fixed time and hoping it works. There is still a timeout on the command that prevents that the test runs forever. I'll push a new version.

PR #156, opened 2026-04-22
docs: Remove Arch Linux from the list of supported systems

sylv-io:
Arch Linux is a bleeding-edge, rolling-release distribution. Tan and I use it for development and always try to resolve any new issues that arise. However, I don't believe we should claim that it's officially supported.


mynetz (approved):
You are right. Agreed. I followed our issues when writing this, but doesn't make sense.

PR #140, opened 2026-04-15
openbmc: srcrev bump ad8d54b743..4bcd6c002b (and its resulting related changes)

sinuscosinustan:
Bump the OpenBMC reference to commit 4bcd6c002b, which also updates Yocto to `wrynose`.

- Pulls in upstream changes across multiple platforms and recipes for the `wrynose` Yocto release.
- Patches refreshed to apply cleanly on top of the new upstream base.
- Fixed source directory assignments: `wrynose` now requires sources to reside in `UNPACKDIR` instead of `WORKDIR`; git-sourced recipes no longer support re-assigning `S`.
- The FIT image generation moved from a custom implementation to the upstream `linux-yocto-fitimage` recipe following recent Yocto recipe updates.
- The GXP validates the U-Boot signature stored in a separate `uboot-sig` partition. If this partition is not updated alongside U-Boot, SRoT validation fails and the iLO bricks. The `uboot-sig` image is now included in `static.mtd.tar` and declared as an optional image in `phosphor-bmc-code-mgmt`.
- **IMPORTANT**: A one-time update using `static.mtd.all.tar` is required when applying this change from an older build, as the previous code does not copy `uboot-sig` to the initramfs. Using `static.mtd.tar` before this commit is applied will brick the iLO.
- Upstream webui-vue migrated from Vue-CLI to Vite. `NODE_ENV` is deprecated in Vite; the dotenv file is now copied to `.env.production` to restore working theme/gzip compression.
- Removed the dumps-page fix patch, which was merged upstream as `5c4f461`.
- Added a `go-task` helper to automate future OpenBMC srcrev bumps.

PR #158, opened 2026-04-23
Prevent own user and root user deletion

### LINKED ISSUE #157: OpenBMC: Removing root user / own user via Redfish possible
# Issue Summary

`bmcweb` does not validate on `DELETE /redfish/v1/AccountService/Accounts/<str>/` if the current user is about to be deleted, and if an error is being returned by the user manager, then it'll always return `resourceNotfound` instead of distinguishing between the error codes. 

## Expected behavior

- Own user can't be deleted (and especially the root user)
- Proper error being returned by bmcweb when user manager returns an error

## Current behavior

- Own user can be deleted (even root)
- error code always returns HTTP 404

## Workaround

None. It's an upstream bug in `bmcweb`

## Build environment

```
Selected build target: canopy-qemu
```

sinuscosinustan:
This PR adds three patches. Two for bmcweb, one for user-manager.

user-manager:
  - Prevent uid 0 deletion

bmcweb:
  - Prevent deletion of own user account
  - Return an actual error instead of using 404 not found for every error message

Closes #157

### DIFF
--- a/meta-canopy/recipes-phosphor/interfaces/bmcweb/0001-account-service-prevent-deletion-of-own-account.patch
+++ b/meta-canopy/recipes-phosphor/interfaces/bmcweb/0001-account-service-prevent-deletion-of-own-account.patch
@@ -1,0 +1,44 @@
+From 43374d7bfa8c79536b83ff24f5f1b94b3e3369ac Mon Sep 17 00:00:00 2001
+From: Tan Siewert <tan.siewert@9elements.com>
+Date: Wed, 22 Apr 2026 18:20:04 +0200
+Subject: [PATCH 1/2] account-service: prevent deletion of own account
+
+Deleting the own account via the Redfish AccountService could leave the
+system in an inconsistent or unmanageable state.
+
+Reject such requests with an InsufficientPrivilege response before any
+D-Bus interaction takes place.
+
+Tested: Deleting own user fails with InsufficientPrivileges
+        successfully.
+
+Upstream-Status: Pending
+Change-Id: I9a519593a43ce14b7c136d16c264dc753a7204eb
+Signed-off-by: Tan Siewert <tan.siewert@9elements.com>
+[Tan: modified patch because the "shorter type aliases" commit is missing]
+---
+ redfish-core/lib/account_service.hpp | 8 ++++++++
+ 1 file changed, 8 insertions(+)
+
+diff --git a/redfish-core/lib/account_service.hpp b/redfish-core/lib/account_service.hpp
+index a91d707f..f8226fb5 100644
+--- a/redfish-core/lib/account_service.hpp
++++ b/redfish-core/lib/account_service.hpp
+@@ -2172,6 +2172,14 @@ inline void handleAccountDelete(
+         messages::resourceNotFound(asyncResp->res, "ManagerAccount", username);
+         return;
+     }
++
++    // Do not allow the removal of the own user
++    if (username == req.session->username)
++    {
++        messages::insufficientPrivilege(asyncResp->res);
++        return;
++    }
++
+     sdbusplus::message::object_path tempObjPath(rootUserDbusPath);
+     tempObjPath /= username;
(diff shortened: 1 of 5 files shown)


sinuscosinustan:
Blocked by #140

meta-canopy/recipes-phosphor/interfaces/bmcweb/0001-account-service-prevent-deletion-of-own-account.patch
```diff
@@ -0,0 +1,44 @@
...
+index a91d707f..f8226fb5 100644
+--- a/redfish-core/lib/account_service.hpp
++++ b/redfish-core/lib/account_service.hpp
+@@ -2172,6 +2172,14 @@ inline void handleAccountDelete(
+         messages::resourceNotFound(asyncResp->res, "ManagerAccount", username);
+         return;
+     }
++
++    // Do not allow the removal of the own user
++    if (username == req.session->username)
```
walterchris, line 33: I know that openBMC is very strict on no business logic in bmcweb. Not sure if that applies here - could this also be handled in the underlying layer?

meta-canopy/recipes-phosphor/users/phosphor-user-manager/0001-user-mgr-prevent-deletion-of-user-with-uid-0.patch
```diff
@@ -0,0 +1,80 @@
...
++}
++
+ void UserMgr::throwForMaxGrpUserCount(
+     const std::vector<std::string>& groupNames)
+ {
+@@ -540,6 +552,7 @@ void UserMgr::deleteUserImpl(const std::string& userName)
+ void UserMgr::deleteUser(std::string userName)
+ {
+     throwForUserDoesNotExist(userName);
++    throwForUidZero(userName);
```
walterchris, line 55: Again - not sure:

in phosphor-user-manager we have `ENABLE_ROOT_USER_MGMT` - Shouldn't this guard this call?

And a thought so: Isn't it generally allowed to remote the root user if we created another user with the same privileges? Is that possible?

walterchris:
It's mostly meant as quetions/thoughts - and not necessary blocking.

walterchris (approved):

meta-canopy/recipes-phosphor/interfaces/bmcweb/0001-account-service-prevent-deletion-of-own-account.patch
sinuscosinustan: The condition was inspired by a similar condition in [handleAccountPatch](https://github.com/openbmc/bmcweb/blob/af44b2bdfb586793124a30e3ad6d077446e5f7dd/redfish-core/lib/account_service.hpp#L2250).

And unfortunately no. `phosphor-user-manager` does not know anything about the requester user. Only bmcweb does here.

meta-canopy/recipes-phosphor/users/phosphor-user-manager/0001-user-mgr-prevent-deletion-of-user-with-uid-0.patch
sinuscosinustan: Disabling `ENABLE_ROOT_USER_MGMT` would mean that root can't be configured via Redfish at all. Yes, it'd not be listed anymore, but it'd also mean that we don't have an option to manage the user anymore.

(fun fact: `root_user_mgmt` is not being used by any upstream platform at this time)

> Isn't it generally allowed to remote the root user if we created another user with the same privileges? Is that possible?

The problem is that `root` is some-what a special user. Deleting the user with UID 0 (which is root in our case) already fails internally, because it owns PID 1 and all system-relevant files on the system, but it still deletes the user from `passwd` and `shadow`.
An alternative would be to create an alternative user who's not UID 0 but still has the privileges like UID 0, though that'd be really special.

Issue #109, opened 2026-03-20
Canopy: Improve image signing and allow usage of external PKI/HSM


mynetz:
# Feature Summary

Images for the HPE ProLiant Gen11 systems need a key to sign the U-Boot images (provided via `HPE_SIGNING_KEY`) . OpenBMC has other places where a key can be used. Given by `SIGNING_KEY` (or `SIGNING_PUBLIC_KEY`). These are unconnected settings for now.

It would be ideal that in normal configurations only one key needs to provided. With the option to provide dedicated keys for the different use-case, if a use wants this.

E.g. set `HPE_SIGNING_KEY` to `SIGNING_KEY` per default.

Also, to have builds that are signed out-of-the-box with a secure key, the signing needs to support HSM (or other external PKI) usage.

This issue is only concerned with the key usage and signing via existing features. In case the research indicates that there are open issues where signing/verification can be improved, raise new issues that address this - ideally using the infrastructure developed with the work on this issue.

## Tasks

- Investigate and document the locations which use a key
- Propose a unified approach that allows overwrites
- Investigate usage of external PKI/HSM via PKCS#11
- Propose and implement a solution for external key storage

## References

- Software HSM to have a PKCS#11 interface that serves local filesystem keys
    - https://github.com/softhsm/SoftHSMv2
    - Only for testing. Real world usage should use a physical HSM.

mynetz:
It seems that the support for HSMs this is already there see [meta-openembedded/meta-oe/classes/signing.bbclass](https://github.com/openbmc/openbmc/blob/ad8d54b743b6118f406010d72bf76d3fcb13d2e9/meta-openembedded/meta-oe/classes/signing.bbclass)

As far as I understand, if properly used/enabled it would always use PKCS#11 but with SoftHSM for local keys.

mynetz:
I investigates this last week further. This will be a multi-part effort. There are the main areas.

1. Signing of U-Boot image via HPE_SIGNING_KEY
2. Signing of OpenBMC partitions via SIGNING_KEY
3. FIT image signing

This issue is only concerned with `[HPE_]SIGNING_KEY`. I have findings or the FIT image that will result in an additional image. See below.

## `[HPE_]SIGNING_KEY`

Right now we only use the `HPE_SIGNING_KEY` and `SIGNING_KEY` is ignored. This leads the OBMC build to use the default private key in the repo - which is considered unsafe.

[...]

PR #172, opened 2026-04-29
meta-hpe: u-boot: Enable FIT signature validation

### LINKED ISSUE #151: HPE ProLiant Gen11: FIT image signing indicated in config but has no effect
# Issue Summary

We set `FITIMAGE_SIGN=1` which has no effect. Since we also select `kernel-fitimage` class. `FITIMAGE_SIGN` is used by `fitimage.bbclass`. The proper value for `kernel-fitimage` would be `UBOOT_SIGN_ENABLE`. This seems to be a mismatch between OpenBMCs Poky class `kernel-fitimage.bbclass` and what Yocto/OE provide via `fitimage.bbclass`
    https://github.com/canopybmc/canopybmc/blob/0c793cd1ce91db7d6e22e1c37457d4bd653598ba/meta-canopy/meta-hpe/meta-gxp/conf/machine/include/hpe_soc.inc#L4-L6

I recall that @sylv-io said that the FIT image validation takes a lot of time since we can't use the crypto extension of the GXP. We should discuss (in the other issue), if we want the signing or not. Right now the config seems to suggest it should be on but isn't.

We either need to use the proper values from `kernel-fitimage` or use `fitimage`. In case the URI signing approach should be used we want to use `fitimage`. It has already a reference. However, I didn't check the usage for it and also don't know the functional impact for the FIT image.

See also: https://github.com/canopybmc/canopybmc/issues/109#issuecomment-4279010156

## Expected behavior

That depends on the discussion and probably some measurements. Ideally we want to have a signed (and validated FIT image). Although that may depends on the intended boot performance. We can maybe provide the user with an overwrite.

However, setting a flag in the default config that suggests it is on, but with no effect is considered a configuration bug.

## Current behavior

No signing done but indicated.

## Workaround

I did not investigate.

sinuscosinustan:
Inherit the uboot-sign bbclass for the HPE GXP platform and implement a custom do_uboot_assemble_fitimage task to inject the RSA public key into the U-Boot DTB at build time (mainly because of the pubkey generation which is not separate in uboot-sign.)

The signing key is the HPE signing key that's being used for the SRoT validation will be re-used.

Note: This implementation does *not* provide PKCS#11 support. There are still some open issues left that need to be addressed first to provide PKCS#11 support.


meta-canopy/meta-hpe/meta-gxp/recipes-bsp/u-boot/u-boot_%.bbappend
```diff
@@ -16,6 +19,51 @@ SRC_URI += "file://0008-board-hpe-gxp-configure-PCIe-device-ID-in-board_init.pat
...
+HPE_OSFCI_SIGNING_KEY = "hpe_osfci_private_key.pem"
+HPE_SIGNING_KEY ?= "${HPE_OSFCI_SIGNING_KEY}"
+UBOOT_SIGN_KEYDIR = "${B}/hpe-fit-keys"
+
+do_uboot_assemble_fitimage() {
+    cd "${B}"
+    install -d "${B}/hpe-fit-keys"
+    install -m 0600 "${UNPACKDIR}/${HPE_SIGNING_KEY}" "${B}/hpe-fit-keys/hpe_fit.key"
+
+    openssl req -new -x509 -days 36500 -subj "/" \
```
ChriMarMe: Is there a valid reason to have a lifetime of 10 years? 

My intuition tells me this shall be 2 years max, but maybe there is a very good reason for 10 which you can explain.

mynetz:
I didn't manage to fully work on this. I will need to rebase the stuff I tested for canopybmc/canopybmc#109 and see. I didn't manage to do this this week. Will come back to this.

I wonder why there is a step to drive a key based on the HPE key variable. Need to get my head back into this...

meta-canopy/meta-hpe/meta-gxp/recipes-bsp/u-boot/u-boot_%.bbappend
sinuscosinustan: My idea with the 10 years lifetime was that the firmware can still be used in e.g. 8 years in case there are no new updates for the platform, and given that the user _might_ not have its own HPE key (because of refurbishment, for example), it'd mean the firmware may not be bootable after its expiry.

walterchris: What happens if the lifetime is expired? Does the platform not boot anymore? Should we allow to boot the platform with expired certificates and somewhat log this?

mynetz: Two things here:
1. Haha! Nice one. I got fooled myself with the 10 years. Collective illusion. The code that @sinuscosinustan wrote does not set *10* years. It is *100* years. The flag sets days. 3650 is 10 years. 36500 is 100 years. Considerning the lifetime: I am confident HPE builds reliable HW, but I guess even they won't see the 100 year mark
2. More importantly the cert doesn't matter. At least this is how I read they documentation. Signing is done with the private key and U-Boot (mkimage) puts the components of the public key in the DTB. The cert seems to be the carrier of the public key.
See: https://docs.u-boot.org/en/latest/usage/fit/signature.html#public-key-storage

Since the cert is never saved anywhere it has to be an ephermal component otherwise we would be screwed anyways... regardless of time.

EDIT: Just for the protocol: RSA keys have no life-time. Only the X509 certificates have. Since the image validation is public/private key based on RSA. It should not matter.

sinuscosinustan: Oh shoot 🙈 Glad that you've noticed it!

So given that the expiration date seems to be unused, I guess we can reduce it to like a few days and call it a day?
I was under the impression that it uses the full X.509 cert and not just the RSA key, but you're right with the docs you've attached

mynetz: @sinuscosinustan I'd remove the flag. The default is 30 days. Which is fine for the purpose.

mynetz:
With regards for signing with a PKI/HSM. It seems that is already supported but may need slightly different options set to mkimage.

@sinuscosinustan I am currently reworking my PoC for #109. I plan to create a PR soon. My intend would be to base this on the work you did here. Then we can merge this change and put the other stuff on top. The changes in this current PR seems self-contained enough.

meta-canopy/meta-hpe/meta-gxp/recipes-bsp/u-boot/u-boot_%.bbappend
mynetz:
```suggestion
    openssl req -new -x509 -subj "/" \
```

meta-canopy/meta-hpe/meta-gxp/recipes-bsp/u-boot/u-boot_%.bbappend
```diff
@@ -16,6 +19,51 @@ SRC_URI += "file://0008-board-hpe-gxp-configure-PCIe-device-ID-in-board_init.pat
 SRC_URI += "file://0009-board-hpe-gxp-add-baseboard-PCA-VPD-to-sysinfo-and-.patch"
 ERROR_QA:remove = "patch-status"
 
+# Override HPE_SIGNING_KEY in local.conf for production builds.
+HPE_OSFCI_SIGNING_KEY = "hpe_osfci_private_key.pem"
+HPE_SIGNING_KEY ?= "${HPE_OSFCI_SIGNING_KEY}"
+UBOOT_SIGN_KEYDIR = "${B}/hpe-fit-keys"
```
mynetz:
```suggestion
UBOOT_SIGN_KEYDIR = "${B}/${UBOOT_SIGN_KEYNAME}-keys"
```

meta-canopy/meta-hpe/meta-gxp/recipes-bsp/u-boot/u-boot_%.bbappend
```diff
@@ -16,6 +19,51 @@ SRC_URI += "file://0008-board-hpe-gxp-configure-PCIe-device-ID-in-board_init.pat
...
+
+do_uboot_assemble_fitimage() {
+    cd "${B}"
+    install -d "${B}/hpe-fit-keys"
```
mynetz:
```suggestion
    install -d "${B}/${UBOOT_SIGN_KEYNAME}-keys"
```

meta-canopy/meta-hpe/meta-gxp/recipes-bsp/u-boot/u-boot_%.bbappend
```diff
@@ -16,6 +19,51 @@ SRC_URI += "file://0008-board-hpe-gxp-configure-PCIe-device-ID-in-board_init.pat
...
+    install -m 0600 "${UNPACKDIR}/${HPE_SIGNING_KEY}" "${B}/hpe-fit-keys/hpe_fit.key"
```
mynetz:
```suggestion
    install -m 0600 "${UNPACKDIR}/${HPE_SIGNING_KEY}" "${B}/${UBOOT_SIGN_KEYNAME}-keys/${UBOOT_SIGN_KEYNAME}.key"
```

meta-canopy/meta-hpe/meta-gxp/recipes-bsp/u-boot/u-boot_%.bbappend
```diff
@@ -16,6 +19,51 @@ SRC_URI += "file://0008-board-hpe-gxp-configure-PCIe-device-ID-in-board_init.pat
...
+
+    openssl req -new -x509 -days 36500 -subj "/" \
+        -key  "${B}/hpe-fit-keys/hpe_fit.key" \
+        -out  "${B}/hpe-fit-keys/hpe_fit.crt"
```
mynetz:
```suggestion
        -key  "${B}/${UBOOT_SIGN_KEYNAME}-keys/${UBOOT_SIGN_KEYNAME}.key" \
        -out  "${B}/${UBOOT_SIGN_KEYNAME}-keys/${UBOOT_SIGN_KEYNAME}.crt"
```

mynetz (changes requested):
I set this to "request change" for the `days` part in the openssl command.

The other proposed changes are just to reduce the duplication of the same string literal that is also set by a variable name already. But that's just my taste and could be ignored ;)

sinuscosinustan:
Failing FWCI tests are unrelated. I've contacted them internally.

mynetz (approved):

meta-canopy/meta-hpe/meta-gxp/recipes-kernel/linux/linux-yocto-fitimage.bbappend
```diff
@@ -10,15 +10,15 @@ SRC_URI += "file://${HPE_SIGNING_KEY}"
...
 
 do_prepare_fit_keys() {
-    install -d "${B}/hpe-fit-keys"
-    install -m 0600 "${UNPACKDIR}/${HPE_SIGNING_KEY}" "${B}/hpe-fit-keys/hpe_fit.key"
-    openssl req -new -x509 -days 36500 -subj "/" \
-        -key "${B}/hpe-fit-keys/hpe_fit.key" \
-        -out "${B}/hpe-fit-keys/hpe_fit.crt"
+    install -d "${B}/${UBOOT_SIGN_KEYNAME}-keys"
+    install -m 0600 "${UNPACKDIR}/${HPE_SIGNING_KEY}" "${B}/${UBOOT_SIGN_KEYNAME}-keys/${UBOOT_SIGN_KEYNAME}.key"
+    openssl req -new -x509 -subj "/" \
```
mynetz, line 19: Nice. I was just realizing that the previous parts where only doing the validation (and embedding of the pubkey).
This is the other side for the actual signing of the image. Seems no-one looked to deep into 69e4ac41caf39a0bac9b2b437128e8225835912b ;)

ChriMarMe: I just noticed the 36500 days thingy. Without knowing anything about the context, I wasn't confident enough to look further into the change, because I have no clue about the process.

PR #173, opened 2026-04-30
meta-canopy: kernel: Move to 6.18.25 stable release

sinuscosinustan:
Apply the upstream security and bug fixes.
Fixes CVE-2026-31431 too.


sylv-io (approved):
I'm always happy when kernel version bumping doesn't require rebasing. 😁

PR #177, opened 2026-05-05
ci: add weekly openbmc bump job

sinuscosinustan:
Add an automatic CI job that updates OpenBMC on the first day of the week at 8 AM UTC using the Taskfile that was added with commit 5addfc33.

The job will push a commit reference update to the same branch / PR, ensuring that no stale updates will be left open as PR.

### DIFF
--- a/.github/workflows/update-openbmc.yml
+++ b/.github/workflows/update-openbmc.yml
@@ -1,0 +1,56 @@
+name: Update OpenBMC reference
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 8 * * 1'
+
+jobs:
+  update-openbmc:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Install go-task
+        uses: arduino/setup-task@v2
+        with:
+          version: 3.x
+
+      - name: Set up branch
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git branch -D update_openbmc_action 2>/dev/null || true
+          git fetch origin
+          git branch update_openbmc_action
+          git checkout update_openbmc_action
+
+      - name: Run obmc-bump
+        run: go-task -y obmc-bump
+
+      - name: Push branch
+        run: git push origin update_openbmc_action --force
+
+      - name: Create or update PR
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          BODY_FILE=$(mktemp)
+          cat > "$BODY_FILE" <<'BODYEOF'
+          Automated changes by the update-openbmc GitHub Action.
+
+          ```
+          BODYEOF
+
+          git log -1 --pretty=%B >> "$BODY_FILE"
+          echo "```" >> "$BODY_FILE"
+
+          gh pr create \
+            --head update_openbmc_action \
+            --title "openbmc: update ref" \
+            --body-file "$BODY_FILE" \
+          || gh pr edit update_openbmc_action \
+            --title "openbmc: update ref" \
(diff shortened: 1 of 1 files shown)


mynetz:
I am wondering, can't we use dependabot for this? I only briefly went through the change. Though it creates a commit with some additional information on the changes based on a commit ref. This looks similar for dependabot.

```
---
version: 2
updates:
  - package-ecosystem: 'gitsubmodule'
    directory: '/'
    schedule:
      interval: 'weekly'
```

In case we want to follow the exact time as the PR:
```
---
version: 2
updates:
  - package-ecosystem: 'gitsubmodule'
    directory: '/'
    schedule:
      interval: 'weekly'
      day: "monday"
      time: "08:00"
      timezone: "Etc/UTC"
```

In one of my projects the default commit message (for the linux submodule) then looks like:
`chore: bump linux from 444b39e to c2d104a`

sylv-io:
@walterchris, create a similar PR #44 in which [I expressed my concerns](https://github.com/canopybmc/canopybmc/pull/44#issuecomment-3920268544) about the lack of automated testing when it was proposed. However, this is no longer the case, so introducing it now makes sense.

I'm assigning him as the reviewer for this PR. :+1:

sinuscosinustan:
@mynetz Oh, true! Tbh, I never used `dependabot` and always disabled it given that I almost exclusively used GitLab for the past 6 1/2 years. Would be worth a shot, I guess?

Though, it'd be interesting if it includes all the commits and authors too, like we do it in the Taskfile right now.

.github/workflows/update-openbmc.yml
```diff
@@ -0,0 +1,56 @@
...
+          git branch -D update_openbmc_action 2>/dev/null || true
+          git fetch origin
+          git branch update_openbmc_action
+          git checkout update_openbmc_action
+
+      - name: Run obmc-bump
+        run: go-task -y obmc-bump
+
+      - name: Push branch
+        run: git push origin update_openbmc_action --force
```
walterchris, line 34: Are we really want to always force push on a single branch? Shouldn't we have dated branches?

.github/workflows/update-openbmc.yml
```diff
@@ -0,0 +1,56 @@
...
+          Automated changes by the update-openbmc GitHub Action.
+
+          ```
+          BODYEOF
+
+          git log -1 --pretty=%B >> "$BODY_FILE"
+          echo "```" >> "$BODY_FILE"
+
+          gh pr create \
+            --head update_openbmc_action \
```
walterchris, line 51: I think I'm more in favor of having dated branches and new PRs every week rather than pushing to the same PR over and over again (imagine the PR is not getting merged).

.github/workflows/update-openbmc.yml
sinuscosinustan: In my opinion it should be one branch since it's a rolling update process. Having dated branches could bloat the active branches and possibly also cause overlapping PRs in case we don't merge one of the bumps.

.github/workflows/update-openbmc.yml
sinuscosinustan: This is exactly the goal I was looking for: No overlapping PRs, one branch, and the open PR updates itself when a (force) push occurs. If a new revision comes up and the previous PR was already merged, then it'll open a new PR instead of re-opening it.

The problem I see with dated branches and possibly overlapping PRs is that it would add complexity in the review process if a newer revision gets merged (i.e. we need logic to automatically close PRs of older revisions and cleaning up the stale branches) which has no real benefit for us.

walterchris (approved):
I don't want to block this because of the naming schema. I want this to happen so let's get it in.

mynetz:
To sum up a discussion from the weekly sync (w/ @sinuscosinustan, @walterchris, @pmaione and @mynetz).

Dependabot does not work with the vision. We want to extend the pipeline in the future to also evaluate patches that we may drop automatically in case they appear upstream. That requires custom tooling anyhow.

PR #178, opened 2026-05-06
HPE Proliant Gen11: Move to OpenBMC kernel tree + tune optimisations

sinuscosinustan:
This PR migrates from kernel.org's tree to the OpenBMC tree, which is required for upstreaming our work to OpenBMC (they require the use of the tree that is managed on github.com/openbmc.), and fixes the tunes from ARMv6-novfp to Cortex-A9 without VFP (ARMv7-a).


sinuscosinustan:
@sylv-io @mynetz I'd like to have both your opinions on the tune and kernel changes that I made here.

From a testing perspective it looks like everything works, and I can see some time improvements on multi-threaded apps as TLS (Thread local-storage) is now available.

As described in the commit message for moving away from `linux-stable` to `linux-hpe` (which uses openbmc/linux now), this is a requirement to be able to start upstreaming to OpenBMC itself.

meta-canopy/meta-hpe/meta-gxp/recipes-kernel/linux/linux-hpe/common/defconfig
```diff
@@ -37,6 +37,7 @@ CONFIG_MTD_SPI_NOR=y
 CONFIG_OF_OVERLAY=y
 CONFIG_BLK_DEV_LOOP=y
 CONFIG_BLK_DEV_RAM=y
+CONFIG_BLK_DEV_NBD=y
```
mynetz, line 40: Where do we need the network block stuff?

meta-canopy/meta-hpe/meta-gxp/recipes-kernel/linux/linux-hpe/common/openbmc-common.cfg
```diff
@@ -0,0 +1,21 @@
...
+CONFIG_SCHED_STACK_END_CHECK=y
+CONFIG_FORTIFY_SOURCE=y
+CONFIG_HARDENED_USERCOPY=y
+CONFIG_SLAB_FREELIST_RANDOM=y
+CONFIG_SLAB_FREELIST_HARDENED=y
+CONFIG_SYN_COOKIES=y
+CONFIG_IP_MULTICAST=y
+CONFIG_NETFILTER=y
+CONFIG_NETCONSOLE=y
+CONFIG_DEBUG_FS=y
```
mynetz, line 16: I am unsure to enable debugfs per default (and always). I will need context.

I did not review how this applies to OBMC or is handled in general in that distribution. Purely judging by an embedded (or single purpose) system, I'd avoid having debug (and other development focused) extensions enabled in a final/production system.

In case this is the norm for OBMC I won't block it. However I'd argue to generally have a look at a later date. Specifically on implications by that on size, performance or security.

mynetz (commented):
Purely source based review. See comments. No approve or deny. Technically it all seems fine and no hard block.

meta-canopy/meta-hpe/meta-gxp/recipes-kernel/linux/linux-hpe/common/defconfig
sinuscosinustan: NBD is required for virtual media support

meta-canopy/meta-hpe/meta-gxp/conf/machine/include/hpe_soc.inc
```diff
@@ -1,4 +1,4 @@
-PREFERRED_PROVIDER_virtual/kernel ?= "linux-stable"
+PREFERRED_PROVIDER_virtual/kernel ?= "linux-hpe"
```
sylv-io, line 1: I understand that this is a pattern used in obmc, but I am not sure whether we should adopt it. We aim to align closely with the latest upstream Linux version, using patches as an intermediate state. This is why I introduced `linux-stable` as the basis for our stable releases.
Since we are not going to maintain separate branches based on SoC vendors in canopy, I don't see the benefit of defining separate providers here.

Therefore, I propose that we leave things as they are and switch to the obmc kernel provider as soon as it is available upstream.

meta-canopy/meta-hpe/meta-gxp/recipes-kernel/linux/linux-hpe/common/openbmc-common.cfg
```diff
@@ -0,0 +1,21 @@
+CONFIG_SOFTLOCKUP_DETECTOR=y
+CONFIG_BOOTPARAM_SOFTLOCKUP_PANIC=y
+CONFIG_BOOTPARAM_HUNG_TASK_PANIC=y
```
sylv-io, line 3: obmc explicitly sets this to =n in their production config. see:
- `meta-phosphor/recipes-kernel/linux/linux-%.bbappend`
- `meta-phosphor/recipes-kernel/linux/production/hung-task.cfg`

meta-canopy/meta-hpe/meta-gxp/conf/machine/include/hpe_soc.inc
sylv-io, line 1: I understand that this is a pattern used in obmc, but I am not sure whether we should adopt it. We aim to align closely with the latest upstream Linux version, using patches as an intermediate state. This is why I introduced linux-stable as the basis for our stable releases.
Since we are not going to maintain separate branches based on SoC vendors in canopy, I don't see the benefit of defining separate providers here.

Therefore, I propose that we leave things as they are and switch to the obmc kernel provider as soon as it is available on upstream obmc.

meta-canopy/meta-hpe/meta-gxp/recipes-kernel/linux/linux-hpe/common/openbmc-common.cfg
```diff
@@ -0,0 +1,21 @@
...
+CONFIG_PANIC_ON_OOPS=y
+CONFIG_PANIC_TIMEOUT=-1
```
sylv-io, line 5: I like that these changes cause the BMC to reboot and recover from any error or corrupted state. However, in the event of a panic, lost uptime will be the only indication in the event of a panic.
Instead, we should look for a solution that generates a critical event log with relevant information about the issue (e.g., OOPS, softlockup.. ) before rebooting.

The kernel provides solutions like ramoops and panic logger, which survive after a restart: https://docs.kernel.org/admin-guide/ramoops.html#ramoops-oops-panic-logger

obmc already has services for it: systemd-pstore, ramoops-monitor and the dreport ramoops plugin for phosphor-debug-collector.

iiuc the only thing missing is enabing pstore:
```
CONFIG_PSTORE=y
CONFIG_PSTORE_RAM=y
CONFIG_PSTORE_CONSOLE=y
CONFIG_PSTORE_PMSG=y
```

and reserving memory dts for ramoops. e.g.:
```
reserved-memory {
	#address-cells = <1>;
	#size-cells = <1>;
	ranges;

	ramoops@5ff00000 {
		compatible = "ramoops";
		reg = <0x5ff00000 0x100000>;
		record-size = <0x20000>;
		console-size = <0x20000>;
		pmsg-size = <0x20000>;
		ftrace-size = <0x0>;
	};
};
```

meta-canopy/meta-hpe/meta-gxp/conf/machine/include/hpe_soc.inc
sylv-io: duplicate, see below. (branch changed during my review)

meta-canopy/meta-hpe/meta-gxp/recipes-kernel/linux/linux-hpe/common/openbmc-common.cfg
sylv-io: we should be able to test it using `sysrq` to trigger a panic.

meta-canopy/meta-hpe/meta-gxp/recipes-kernel/linux/linux-hpe/common/openbmc-common.cfg
sinuscosinustan: Will keep them as-is but set the respective distrooverride when the release channel is configured, so that they are set to `=n`. Makes it also easier to upstream it.

meta-canopy/meta-hpe/meta-gxp/conf/machine/include/hpe_soc.inc
sinuscosinustan: As explained in our recent meeting on 2026-05-18, it is not about branching the kernel for each SoC vendor in git. The goal is to allow moving forward to newer kernel versions without breaking all vendors simultaneously (or even going forward to a newer LTS release), and being compliant with the OpenBMC upstream implementation for now. Getting a different pattern into upstream will be quite some work that should not block us completely.

The Linux dev-6.18 tree is a maintained fork of `stable/linux.git` with some backports and patches that aren't in upstream (yet).

For now, I'd like to get a base into upstream first that complies with it, even if it is "wrong".

meta-canopy/meta-hpe/meta-gxp/recipes-kernel/linux/linux-hpe/common/openbmc-common.cfg
sinuscosinustan: As discussed in the meeting on 2026-05-18, it'll make more sense to divert on this one and disable debugfs on release/production builds (see my comment on Marcello's review.)

meta-canopy/meta-hpe/meta-gxp/recipes-kernel/linux/linux-hpe/common/openbmc-common.cfg
sinuscosinustan: Thanks for the suggestion of adding ramoops! I'll test it and report back.

sinuscosinustan: Works perfectly! Implemented with commit 53a3d14 `meta-hpe: kernel: enable ramoops`.

meta-canopy/meta-hpe/meta-gxp/recipes-kernel/linux/linux-hpe/release/production.cfg
mynetz, line 1: Nice! Okay, this addresses https://github.com/canopybmc/canopybmc/pull/178/changes#r3268069756

mynetz (approved):

sylv-io (approved):
lgtm :)

PR #186, opened 2026-05-21
meta-canopy: bmcweb: add zstd fix

sinuscosinustan:
After the regular OpenBMC bump in commit 9da52a04, bmcweb got bumped to a commit that reduces the memory usage by not copying the data again. Unfortunately, this resulted in all requests to persistent files on the system (e.g. Web UI) to return zero data if `accept-encoding: zstd` was set by the client because the FD got closed early / the FileBody got destroyed after the body content check (string emplace).

This adds a fix that was written after a discussion on Discord.


ChriMarMe (approved):
LGTM from a code and commit msg pov.

PR #199, opened 2026-06-15
ci: use runner-provided sstate/download cache and hashserve db

sinuscosinustan:
Instead of using our own sstate and download cache (which might get erased), use the central one provided by the runner itself.

This also fixes the issue where the build process was not using a central hashserve db, causing the sstate to be rebuilt every time regardless of its state.

The mentioned link refers to the fix in the runner configuration.

Link: https://github.com/9elements/infrastructure/pull/235

### DIFF
--- a/.github/actions/build/action.yml
+++ b/.github/actions/build/action.yml
@@ -11,12 +11,6 @@ inputs:
 runs:
   using: 'composite'
   steps:
-    - name: Configure bitbake shared folders
-      shell: bash
-      run: |
-        mkdir -p /home/runner/data/bitbake.downloads
-        mkdir -p /home/runner/data/bitbake.sstate
-
     - name: Init git submodules
       shell: bash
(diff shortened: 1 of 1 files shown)


mynetz (approved):
Seems fine to me. My understanding might be wrong, but this change should ensure that subsequent builds are faster. However, this seems to be the case already for the same branch.
I guess the intend is to make that work on several branches.

Since it generally works and does not break anything I tend to approve it.
- First build: https://github.com/canopybmc/canopybmc/actions/runs/27540001348
- Second build: https://github.com/canopybmc/canopybmc/actions/runs/27695589444

Now, it seems to me that this does not fully test the full intend. Since it seems that rebuilds on the same branch were already faster in the past. The aim with this change is to have this on different branches/builds. Which is harder to test now.
I would approve this now and in case the goal isn't reached, we revise this.

Issue #198, opened 2026-06-12
HPE Proliant Gen11: dev channel rofs out of space


sinuscosinustan:
# Issue Summary

Builds with the dev channel fail due to space constraints on `rofs`:
```
ERROR: obmc-phosphor-image-1.0-r0 do_generate_static: Image size error: '/home/tan/git/github.com/canopybmc/canopybmc/build/hpe-proliant-g11/tmp/work/hpe_proliant_g11-openbmc-linux-gnueabi/obmc-phosphor-image/1.0/deploy-obmc-phosphor-image-image-complete/obmc-phosphor-image-hpe-proliant-g11.squashfs-xz' is 25292800 bytes, exceeds allowed 25214976 bytes (over by 77824 bytes)
ERROR: Logfile of failure stored in: /home/tan/git/github.com/canopybmc/canopybmc/build/hpe-proliant-g11/tmp/work/hpe_proliant_g11-openbmc-linux-gnueabi/obmc-phosphor-image/1.0/temp/log.do_generate_static.1040599
ERROR: Task (/home/tan/git/github.com/canopybmc/canopybmc/openbmc/meta-phosphor/recipes-phosphor/images/obmc-phosphor-image.bb:do_generate_static) failed with exit code '1'
```

## Expected behavior

It builds.

## Current behavior

It doesn't build.

## Workaround

Use the release channel.

## Build environment

```
Selected build target: hpe-proliant-gen11
```

sinuscosinustan:
The issue ends up in reworking the whole fan sensor stuff because it is way too over-engineered and uses two different approaches that conflict with each other.

sinuscosinustan:
The fan sensor rework will be moved to the next release. We've got enough space with the multi-call binary backport for now. The fan sensor rework will be tracked in #207

PR #205, opened 2026-06-24
meta-hpe: dbus-sensors: backport multi-call binary commit

sinuscosinustan:
Backport a still pending CR from Google to convert dbus-sensors into a multi-call binary which reduces the storage usage by about 1.4MiB in squashfs (i.e. compressed).

This will free a lot of space that we need for more tooling.

Relates to #198


meta-canopy/meta-hpe/meta-common/recipes-phosphor/sensors/dbus-sensors/0001-reduce-image-size-by-moving-to-Multi-call-Binary.patch
```diff
@@ -0,0 +1,1447 @@
...
++#ifdef LEAKDETECTOR
++#include "leakdetector/LeakDetectionManager.hpp"
++#endif
++#ifdef MCTP
++#include "mctp/MCTPReactor.hpp"
++#endif
++#ifdef MCU
++#include "mcu/MCUTempSensor.hpp"
++#endif
++#ifdef NVDIAGPU
```
pmaione, line 221: typo: should be NVIDIAGPU as defined in src/nvidia-gpu/meson.build

But this do not affect our canopy build.

pmaione (approved):
LGTM.

Checked on DL360.

meta-canopy/meta-hpe/meta-common/recipes-phosphor/sensors/dbus-sensors/0001-reduce-image-size-by-moving-to-Multi-call-Binary.patch
sinuscosinustan: Dang! Thanks for this. Would you mind adding this as note in the upstream CR too?

PR #206, opened 2026-06-25
HPE ProLiant Gen11: Add BIOS Update support

### LINKED ISSUE #84: HPE Proliant Gen11: Enable Host BIOS update support
# Feature Summary

Currently, neither the host BIOS nor alternative ROMs can be updated via the BMC as there's no code to flash the ROM to the respective SPIs.  
On the HPE build, there are a various amount of scripts in `phosphor-software-manager` that handle BIOS flashing and SCP EEPROM updates (the latter one is for Ampere only) and image verification if enabled.

Unfortunately, HPE encrypts its BIOS files which, according to @sylv-io, can't be decrypted with OpenBMC as the required key is not available however, the tarball build has a signature validator for the encrypted builds and apparently writes the file as-is to the SPIs (?).

There's also use of 32MB `vrom` partitions that are being written on host boot-up and after a firmware update (writing the content from the host SPI to those `mtd-ram` partitions), though it is unclear to me for what they are being used for (and apparently `vrom` can't be used on Ampere platforms.)

## Tasks

- Clarify what the `vrom` partitions are being used for
- Check if we need to decrypt the BIOS before being able to update it

## References

- `meta-hpe-private/meta-proliant-g11/recipes-phosphor/flash/phosphor-software-manager/hpe-flash-rom.sh` in the tarball

# Note
> This is a tracking item for work that may partially done in different
> repositories. Items are linked as much as possible. Some repositories are
> private and stay private since they are used to verify and clean-up vendor
> (derived) code.

sinuscosinustan:
This series adds support to update the BIOS (both the primary and secondary memory-mapped flash) in a powered-off state.

- 9d97ab5 adds the bash script that's being triggered by `phosphor-bmc-code-mgmt` to update the BIOS
- da56b77 adds a script for us to package the HPE binaries into a tarball that can be used for Redfish, as well as in the Web UI
- 91aed2a adds a fix in the Web UI to allow the user to select the correct target

Closes #84

### DIFF
--- a/meta-canopy/meta-hpe/meta-common/recipes-phosphor/flash/phosphor-software-manager/obmc-flash-host-bios
+++ b/meta-canopy/meta-hpe/meta-common/recipes-phosphor/flash/phosphor-software-manager/obmc-flash-host-bios
@@ -1,0 +1,209 @@
+#!/bin/bash
+
+redfish_log() {
+	# Usage: redfish_log <message> <severity> [key value [key value ...]]
+	local message="$1" severity="$2"
+	shift 2
+	local pairs=() n=0
+	while [ $# -ge 2 ]; do
+		pairs+=("$1" "$2")
+		n=$((n + 1))
+		shift 2
+	done
+	busctl call \
+		xyz.openbmc_project.Logging /xyz/openbmc_project/logging \
+		xyz.openbmc_project.Logging.Create Create "ssa{ss}" \
+		"$message" "$severity" "$n" "${pairs[@]}" 2>/dev/null || true
+}
+
+die() {
+	logger -t obmc-flash-host-bios "Error: $*"
+	redfish_log "BIOS update failed" \
+		"xyz.openbmc_project.Logging.Entry.Level.Error" \
+		"REASON" "$*"
+	exit 1
+}
+
+info() { logger -t obmc-flash-host-bios "$*"; }
+
+BIOS_SIZE=$((32 * 1024 * 1024))
+HPE_FINGERPRINT_MAGIC="--=</"
+HPE_IMAGE_MAGIC="HPIMAGE"
+
+POWER_BUTTON_PATH="/xyz/openbmc_project/chassis/buttons/power"
+POWER_BUTTON_IFACE="xyz.openbmc_project.Chassis.Buttons"
+POWER_CONTROL_SERVICE="$(mapper get-service "$POWER_BUTTON_PATH")"
+
+HOSTSTATE_SVC="xyz.openbmc_project.State.Host"
+HOSTSTATE_PATH="/xyz/openbmc_project/state/host0"
+HOSTSTATE_IFACE="xyz.openbmc_project.State.Host"
+HOSTSTATE_PROP="CurrentHostState"
+HOSTSTATE_OFF="xyz.openbmc_project.State.Host.HostState.Off"
+
+check_host_off() {
+	local state
+	state="$(busctl get-property "$HOSTSTATE_SVC" "$HOSTSTATE_PATH" \
+		"$HOSTSTATE_IFACE" "$HOSTSTATE_PROP")" \
+		|| die "could not read host state from D-Bus"
+	[ "$state" = "s \"$HOSTSTATE_OFF\"" ] \
+		|| die "host must be off before performing BIOS update"
+}
+
+power_mask_set() {
+	if busctl set-property \
+			"$POWER_CONTROL_SERVICE" \
+			"$POWER_BUTTON_PATH" \
+			"$POWER_BUTTON_IFACE" \
+			ButtonMasked b true 2>/dev/null; then
+		info "power button masked"
+	else
+		info "WARNING: could not mask power button"
+	fi
+}
+
+power_mask_clear() {
+	if busctl set-property \
+			"$POWER_CONTROL_SERVICE" \
+			"$POWER_BUTTON_PATH" \
+			"$POWER_BUTTON_IFACE" \
+			ButtonMasked b false 2>/dev/null; then
+		info "power button unmasked"
+	else
+		info "WARNING: could not unmask power button"
+	fi
+}
+
+check_bios_image() {
+	local image_file="$1"
+	local file_size
+
+	[ -r "$image_file" ] || die "can't read BIOS image $image_file"
+
+	file_size="$(stat -c '%s' "$image_file")"
+	[ "$file_size" -ge "$BIOS_SIZE" ] \
+		|| die "invalid BIOS image (smaller than $BIOS_SIZE bytes)"
+}
+
+check_model() {
+	local image_dir="$1"
+	local manifest="$image_dir/MANIFEST"
+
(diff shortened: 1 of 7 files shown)


meta-canopy/scripts/gen-hpe-bios-tar
```diff
@@ -0,0 +1,227 @@
...
+		o) OUTPUT="$OPTARG" ;;
+		h) usage; exit 0 ;;
+		:) die "Option -$OPTARG requires an argument." ;;
+		\?) die "Unknown option: -$OPTARG" ;;
+	esac
+done
+
+[[ -n "$INPUT" ]]	|| die "Missing required argument: -i <input>"
+[[ -n "$MODEL" ]]	|| die "Missing required argument: -m <model>"
+[[ -n "$VERSION" ]] || die "Missing required argument: -v <version>"
```
mynetz, line 143: The version is encoded in the `<T_Identifier>_<version>_<date>[.unencrypted].signed.flash` in multiple locations:
- In the name itself
- In the binary

We may want to support images that are named differently as input, then we can not expect the version there. But it may serve as a probing attempt.

Similar to the model https://github.com/canopybmc/canopybmc/pull/206/changes#r3473279269 bit version could be obtained directly
```bash
strings U63_2.92_05_28_2026.unencrypted.signed.flash | grep 2.92
v2.92 (05/28/2026)
U63 v2.92
ROM_VER_STR=U63 v2.92 (05/28/2026)
...
```

I checked this for the DL320 and DL365 now. The easiest seems to query for the `ROM_MAJOR_VER` and `ROM_MINOR_VER` in the binary.

`-v` could be used to validate or overwrite this, but we should probably just obtain it from the binary and reject if not found.

meta-canopy/scripts/gen-hpe-bios-tar
```diff
@@ -0,0 +1,227 @@
...
+		k) PRIVKEY="$OPTARG" ;;
+		o) OUTPUT="$OPTARG" ;;
+		h) usage; exit 0 ;;
+		:) die "Option -$OPTARG requires an argument." ;;
+		\?) die "Unknown option: -$OPTARG" ;;
+	esac
+done
+
+[[ -n "$INPUT" ]]	|| die "Missing required argument: -i <input>"
+[[ -n "$MODEL" ]]	|| die "Missing required argument: -m <model>"
```
mynetz, line 142: I am undecided. I would like to not block the feature for trivial things, but I also feel this adds a usability burden. Lets discuss this.

It seems we could query this from the rom itself. This could also serve as a way to validate the binary.

The binary seems to contain a list of descriptive "variables":
```bash
strings A55_3.14_05_28_2026.unencrypted.signed.flash | grep ROM_INFO -A10
[ROM_INFO]
ROM_SYSID=A55
ROM_MAJOR_VER=3
ROM_MINOR_VER=14
ROM_PASS=00
ROM_DATE=05/28/2026
ROM_VER_STR=A55 v3.14 (05/28/2026)
```

`ROM_SYSID` is what we would be looking for. This then could be used to lookup the model. Though HPE would need to provide this for us. Compiling this from the list of puplic roms (encrypted) would also work.

Since the model provided is anyhow defined in a list to match the DLxxx name we could also match that via the [A-Z][0-9][0-9] number.

meta-canopy/meta-hpe/meta-common/recipes-phosphor/flash/phosphor-software-manager/obmc-flash-host-bios
```diff
@@ -0,0 +1,209 @@
...
+		flash_image="$(dirname "$image_file")/bios-stripped.bin"
+
+		info "stripping $header_size byte HPE header, writing to $flash_image"
+		dd if="$image_file" of="$flash_image" bs="$header_size" skip=1 2>/dev/null \
+			|| die "failed to strip HPE header from $image_file"
+	else
+		flash_image="$image_file"
+	fi
+
+	for partition in host-prime host-second; do
```
mynetz, line 152: If I read this correctly, this update loop will update the primary flash and then directly the secondary.

I see two issues here:
- The update of the first flash fails and we still loop to write the second, potentially breaking both. Though depends on the workings of flashcp.
- We update both without any run-time validation of the update of the primary flash. Is this normal when done from the BMC side?

mynetz (commented):
I am not sure my comments qualify as blocking. Hence I set them to "comment". Ideally I create a discussion or get some input.

I mostly looked at the usability side. Didn't look at the Yocto and systemd files.

meta-canopy/meta-hpe/meta-common/recipes-phosphor/flash/phosphor-software-manager/obmc-flash-host-bios
sinuscosinustan: > The update of the first flash fails and we still loop to write the second, potentially breaking both. Though depends on the workings of flashcp.

Dang! I had a `set -e` in my first iteration locally and removed it :D The current implementation has this issue, yes.

> We update both without any run-time validation of the update of the primary flash. Is this normal when done from the BMC side?

What exactly do you mean with runtime validation in this case?

For the "we flash both partitions at once", this was partially inspired by the original implementation, but maybe I can check to add the "backup image" functionality to switch to the other image if required.

meta-canopy/scripts/gen-hpe-bios-tar
sinuscosinustan: Grepping through the binary showed that it supports multiple systems with the same binary. Getting a list of the SYSIDs mapped to the models shouldn't be an issue. I'll send them a chat later.

meta-canopy/scripts/gen-hpe-bios-tar
sinuscosinustan: True, this sounds like a good idea! Many thanks for the suggestion.

meta-canopy/meta-hpe/meta-common/recipes-phosphor/flash/phosphor-software-manager/obmc-flash-host-bios
mynetz: > What exactly do you mean with runtime validation in this case?
>
> For the "we flash both partitions at once", this was partially inspired by the original implementation, but maybe I can check to add the "backup image" functionality to switch to the other image if required.

Understood. Yes, this was what I was thinking about. Ideally an update of the flash is done in a way that ensures the host can boot. A/B scheme style or with validation of any sort and then updating the secondary.
Though that all depends on how the HPE platform behaves.

I think in the discussion around this you noted that iLO handles this in a way that it only write the primary and then requires additional action to overwrite the backup/secondary. Though this may not be supported in OpenBMC(? - or redfish?)

sinuscosinustan: See the new PR, there are now two separate flash targets.

sinuscosinustan:
Closed in favour of #210

PR #210, opened 2026-07-01
meta-hpe: enable bios firmware update

### LINKED ISSUE #84: HPE Proliant Gen11: Enable Host BIOS update support

pointbazaar:
This requires a host fw image ready for flashing (without the HPE header which was removed in other script) and packaged as PLDM fw update package with the right compatible and IANA Vendor Id.

Tested: on HPE Proliant G11 D360

Requesting the update

```
curl -k --insecure --user root:0penBmc -H Content-Type:multipart/form-data -X POST -F 'UpdateParameters={"Targets":["/redfish/v1/UpdateService/FirmwareInventory/HostSPIFlash_146"],"@Redfish.OperationApplyTime":"Immediate"};type=application/json' -F 'UpdateFile=@pldm-package-hpe-proliant-g11-d360-host.bin;type=application/octet-stream' https://hpe-dl360-g11-bmc.dut.lab.bs.9e.network:443/redfish/v1/UpdateService/update-multipart
{
  "@odata.id": "/redfish/v1/TaskService/Tasks/1",
  "@odata.type": "#Task.v1_4_3.Task",
  "HidePayload": false,
  "Id": "1",
  "Messages": [
    {
      "@odata.type": "#Message.v1_1_1.Message",
      "Message": "The task with Id '1' has started.",
      "MessageArgs": [
        "1"
      ],
      "MessageId": "TaskEvent.1.0.TaskStarted",
      "MessageSeverity": "OK",
      "Resolution": "None."
    }
  ],
  "Name": "Task 1",
  "Payload": {
    "HttpHeaders": [],
    "HttpOperation": "POST",
    "TargetUri": "/redfish/v1/UpdateService/update-multipart"
  },
  "PercentComplete": 0,
  "StartTime": "2026-06-30T14:54:36+00:00",
  "TaskMonitor": "/redfish/v1/TaskService/TaskMonitors/1",
  "TaskState": "Running",
  "TaskStatus": "OK"
}
```

After completing the update, the new host fw version shows up in fw inventory:
```
{
"@odata.id": "/redfish/v1/UpdateService/FirmwareInventory",
"@odata.type": "#SoftwareInventoryCollection.SoftwareInventoryCollection",
"Members": [
{
  "@odata.id": "/redfish/v1/UpdateService/FirmwareInventory/22ba1852"
},
{
  "@odata.id": "/redfish/v1/UpdateService/FirmwareInventory/HostSPIFlash_9805"
},
{
  "@odata.id": "/redfish/v1/UpdateService/FirmwareInventory/bios_active"
}
],
"Members@odata.count": 3,
"Name": "Software Inventory Collection"
}
```

One remaining issue is that the Redfish Task for this update seems to timeout or get lost after a while, possible because the fw update code is not reporting progress frequently enough. This remains to be investigated but overall the update seems to be working.

The host boots `HPE ProLiant System BIOS U54 v2.92 05/28/2026` so either the wrong flash chip was written or everything went well.

The flash contents are the same as what we wrote.
```
root@hpe-proliant-g11:~# md5sum /dev/mtd/by-name/host-second
9016af89497120bd4434864bbfbab38d  /dev/mtd/by-name/host-second
md5sum /tmp/bios-stripped.bin
9016af89497120bd4434864bbfbab38d  /tmp/bios-stripped.bin
```

Verify by writing an empty image, the contents are as expected
```
root@hpe-proliant-g11:~# md5sum /dev/mtd/by-name/host-second
58f06dd588d8ffb3beb46ada6309436b  /dev/mtd/by-name/host-second
md5sum empty-img-32M.bin
58f06dd588d8ffb3beb46ada6309436b  empty-img-32M.bin
```

see #84

### DIFF
--- a/meta-canopy/meta-hpe/meta-common/recipes-phosphor/configuration/entity-manager/dl360g11_baseboard.json
+++ b/meta-canopy/meta-hpe/meta-common/recipes-phosphor/configuration/entity-manager/dl360g11_baseboard.json
@@ -155,6 +155,7 @@
         },
         {
             "Class": "temp",
+            "IgnoreFailIfHostOff": true,
             "Inputs": [
                 "Die CPU1"
             ],
@@ -189,6 +190,7 @@
         {
             "Class": "temp",
             "InputUnavailableAsFailed": false,
+            "IgnoreFailIfHostOff": true,
             "Inputs": [
                 "Die CPU2"
             ],
--- a/meta-canopy/meta-hpe/meta-common/recipes-phosphor/configuration/entity-manager/dl360g11_baseboard.json
+++ b/meta-canopy/meta-hpe/meta-common/recipes-phosphor/configuration/entity-manager/dl360g11_baseboard.json
@@ -308,6 +308,32 @@
             "Name": "GenericContainPort",
             "PortType": "containing",
             "Type": "Port"
+        },
+        {
+            "FirmwareInfo": {
+                "CompatibleHardware": "com.hpe.Hardware.ProLiantDL360G11.SPI.Host",
+                "VendorIANA": 47196
+            },
+            "MuxOutputs": [
+            ],
+            "Name": "UEFIPrimary",
+            "SPIControllerIndex": 1,
+            "SPIDeviceIndex": 1,
+            "Partition": "host-prime",
+            "Type": "SPIFlash"
+        },
+        {
+            "FirmwareInfo": {
+                "CompatibleHardware": "com.hpe.Hardware.ProLiantDL360G11.SPI.Host",
+                "VendorIANA": 47196
+            },
+            "MuxOutputs": [
+            ],
+            "Name": "UEFISecondary",
+            "SPIControllerIndex": 1,
+            "SPIDeviceIndex": 1,
+            "Partition": "host-second",
+            "Type": "SPIFlash"
         }
     ],
     "Name": "chassis",
--- a/meta-canopy/meta-hpe/meta-common/recipes-phosphor/configuration/entity-manager/dl360g11_baseboard.json
+++ b/meta-canopy/meta-hpe/meta-common/recipes-phosphor/configuration/entity-manager/dl360g11_baseboard.json
@@ -321,19 +321,6 @@
             "Name": "UEFIPrimary",
             "SPIControllerIndex": 1,
             "SPIDeviceIndex": 1,
-            "Partition": "host-prime",
-            "Type": "SPIFlash"
-        },
-        {
-            "FirmwareInfo": {
-                "CompatibleHardware": "com.hpe.Hardware.ProLiantDL360G11.SPI.Host",
-                "VendorIANA": 47196
-            },
-            "MuxOutputs": [
-            ],
-            "Name": "UEFISecondary",
-            "SPIControllerIndex": 1,
-            "SPIDeviceIndex": 1,
             "Partition": "host-second",
             "Type": "SPIFlash"
         }
--- a/meta-canopy/meta-hpe/meta-common/recipes-phosphor/flash/phosphor-software-manager/0001-bios-no-not-require-mux-GPIOs.patch
+++ b/meta-canopy/meta-hpe/meta-common/recipes-phosphor/flash/phosphor-software-manager/0001-bios-no-not-require-mux-GPIOs.patch
@@ -1,0 +1,236 @@
+From 7285b68bd6c1c78ab24d081a3a422921d9a2a86b Mon Sep 17 00:00:00 2001
+From: Alexander Hansen <alexander.hansen@9elements.com>
+Date: Thu, 25 Jun 2026 18:15:05 +0200
+Subject: [PATCH 1/2] bios: no not require mux GPIOs
+
+Not every platform has their SPI flash behind a gpio Mux, we should not
+stop the update process because no lines were configured.
+
+For example HPE ProLiant Gen11 does not have a gpio mux for the SPI
+flash.
+
+Also we should not assume a specific driver. In this patch, the step of
+binding the flash to the driver is skipped if the flash is already
+bound. In case of HPE ProLiant Gen11 it is already bound via devicetree
+and not using aspeed spi smc driver either, so binding that would not
+work since it is using an HPE GXP SoC for BMC.
+
+```
+root@hpe-proliant-g11:~# ls /sys/bus/spi/drivers/spi-nor/
+bind    spi0.0  spi1.0  spi1.1  uevent  unbind
+
+root@hpe-proliant-g11:~# zcat /proc/config.gz | grep -i HPE
+CONFIG_ARCH_HPE=y
+CONFIG_ARCH_HPE_GXP=y
+CONFIG_SOC_HPE=y
+CONFIG_SOC_HPE_GXP=y
+CONFIG_NVMEM_HPE_GXP=y
+```
+
+Tested: on HPE ProLiant Gen11 D360
+
+```
+root@hpe-proliant-g11:~# busctl introspect xyz.openbmc_project.EntityManager /xyz/openbmc_project/inventory/system/chassis/chassis/HostSPIFlash
+...
+xyz.openbmc_project.Configuration.... interface -         -                      -
+.Name                                 property  s         "HostSPIFlash"         emits-change
+.SPIControllerIndex                   property  t         1                      emits-change
+.SPIDeviceIndex                       property  t         1                      emits-change
+.Type                                 property  s         "SPIFlash"             emits-change
+xyz.openbmc_project.Configuration.... interface -         -                      -
+.CompatibleHardware                   property  s         "com.hpe.proliant.g11" emits-change
+.VendorIANA                           property  t         47196                  emits-change
+root@hpe-proliant-g11:~# busctl tree xyz.openbmc_project.Software.BIOS
+`- /xyz
+  `- /xyz/openbmc_project
+    `- /xyz/openbmc_project/software
+      `- /xyz/openbmc_project/software/HostSPIFlash_146
+```
+
+The fw inventory item shows up
+
+```
+curl --silent --show-error --insecure --user root:0penBmc https://hpe-dl360-g11-bmc.dut.lab.bs.9e.network:443/redfish/v1/UpdateService/FirmwareInventory
+{
+  "@odata.id": "/redfish/v1/UpdateService/FirmwareInventory",
+  "@odata.type": "#SoftwareInventoryCollection.SoftwareInventoryCollection",
+  "Members": [
+    {
+      "@odata.id": "/redfish/v1/UpdateService/FirmwareInventory/76399742"
+    },
+    {
+      "@odata.id": "/redfish/v1/UpdateService/FirmwareInventory/HostSPIFlash_146"
+    },
+    {
+      "@odata.id": "/redfish/v1/UpdateService/FirmwareInventory/bios_active"
+    }
+  ],
+  "Members@odata.count": 3,
+  "Name": "Software Inventory Collection"
+}
+```
+
+Starting the update request:
+
+```
+curl -k --insecure --user root:0penBmc -H Content-Type:multipart/form-data -X POST -F 'UpdateParameters={"Targets":["/redfish/v1/UpdateService/FirmwareInventory/HostSPIFlash_146"],"@Redfish.OperationApplyTime":"Immediate"};type=application/json' -F 'UpdateFile=@pldm-package-hpe-proliant-g11-d360-host.bin;type=application/octet-stream' https://hpe-dl360-g11-bmc.dut.lab.bs.9e.network:443/redfish/v1/UpdateService/update-multipart
+{
+  "@odata.id": "/redfish/v1/TaskService/Tasks/1",
+  "@odata.type": "#Task.v1_4_3.Task",
+  "HidePayload": false,
+  "Id": "1",
+  "Messages": [
+    {
+      "@odata.type": "#Message.v1_1_1.Message",
+      "Message": "The task with Id '1' has started.",
+      "MessageArgs": [
+        "1"
+      ],
+      "MessageId": "TaskEvent.1.0.TaskStarted",
+      "MessageSeverity": "OK",
+      "Resolution": "None."
+    }
+  ],
+  "Name": "Task 1",
+  "Payload": {
+    "HttpHeaders": [],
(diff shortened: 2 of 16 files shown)


meta-canopy/meta-hpe/meta-common/recipes-phosphor/configuration/entity-manager/dl360g11_baseboard.json
```diff
@@ -308,6 +308,18 @@
             "Name": "GenericContainPort",
             "PortType": "containing",
             "Type": "Port"
+        },
+	{
+            "FirmwareInfo": {
+                "CompatibleHardware": "com.hpe.proliant.g11",
```
sinuscosinustan: Please be specific with the compatible on that as the BIOS blobs are not inter-compatible.


```suggestion
                "CompatibleHardware": "com.hpe.Hardware.ProLiantDL360G11.SPI.Host",
```

meta-canopy/meta-hpe/meta-common/recipes-phosphor/configuration/entity-manager/dl360g11_baseboard.json
```diff
@@ -308,6 +308,18 @@
...
+                "VendorIANA": 47196
+            },
+            "MuxOutputs": [
+            ],
+            "Name": "HostSPIFlash",
+            "SPIControllerIndex": 1,
+            "SPIDeviceIndex": 1,
```
sinuscosinustan, line 323: that'll only update the secondary partition rn and there's currently no docs on how to switch to `host-secondary` for booting. Consider using `"SPIDeviceIndex": 0` for now. We'll discuss this with HPE.

meta-canopy/meta-hpe/meta-common/recipes-phosphor/flash/phosphor-software-manager/0001-bios-no-not-require-mux-GPIOs.patch
```diff
@@ -0,0 +1,236 @@
...
+  ],
+  "Members@odata.count": 3,
+  "Name": "Software Inventory Collection"
+}
+```
+
+Change-Id: I7e24011890b4e28fbe832746de877078864a41a6
+Signed-off-by: Alexander Hansen <alexander.hansen@9elements.com>
+---
+Upstream-Status: Pending
```
sinuscosinustan:
```suggestion
Upstream-Status: Submitted [Gerrit, https://gerrit.openbmc.org/c/openbmc/phosphor-bmc-code-mgmt/+/91736]
```

sinuscosinustan (changes requested):

meta-canopy/meta-hpe/meta-common/recipes-phosphor/configuration/entity-manager/dl360g11_baseboard.json
pointbazaar: We need to go with `"SpiDeviceIndex": 1`. 
These numbers map directly to spi controller index and device index (chip select)

The problem now is that it looks like `spi1.1` has multiple flash partitions

```
root@hpe-proliant-g11:~# cat /sys/class/spi_master/spi1/spi1.1/mtd/mtd11/name 
host-prime
root@hpe-proliant-g11:~# cat /sys/class/spi_master/spi1/spi1.1/mtd/mtd12/name 
host-second
```

This is not something the upstream software supports yet. Presumably there was no use case for it as there was no A/B scheme.

It would be straightforward to add it, all we would need is probably a `PartitionName` property on the configuration record to support partitioned flash chips.

```
   flash@1 {
    compatible = "jedec,spi-nor";
    reg = <1>;
    partitions {
     compatible = "fixed-partitions";
     #address-cells = <1>;
     #size-cells = <1>;

     host-prime@0 {
      label = "host-prime";
      reg = <0x0 0x2000000>;
     };
     host-second@2000000 {
      label = "host-second";
      reg = <0x2000000 0x2000000>;
     };
    };
   };
  };
 };
```

I will leave this open for now as the problem has not been fixed. Currently the upstream code selects the first mtd partition it finds which may not be the right one and not even sure if its deterministic.

meta-canopy/meta-hpe/meta-common/recipes-phosphor/configuration/entity-manager/dl360g11_baseboard.json
pointbazaar: Thanks, changed it.

meta-canopy/meta-hpe/meta-common/recipes-phosphor/configuration/entity-manager/dl360g11_baseboard.json
pointbazaar: removed this limitation by adding support for configuring which partition to update.

meta-canopy/meta-hpe/meta-common/recipes-phosphor/configuration/entity-manager/0001-schemas-firmware-allow-for-updating-spi-partition.patch
```diff
@@ -0,0 +1,78 @@
...
+diff --git a/schemas/firmware.json b/schemas/firmware.json
+index ba32140..52da6be 100644
+--- a/schemas/firmware.json
++++ b/schemas/firmware.json
+@@ -95,6 +95,10 @@
+                     "description": "The index of the SPI device connected to that controller",
+                     "type": "number"
+                 },
++                "Partition": {
++                    "description": "The name of the partition to update. Leave empty if there is only a single partiion without a name.",
```
mynetz:
```suggestion
+                    "description": "The name of the partition to update. Leave empty if there is only a single partition without a name.",
```

meta-canopy/scripts/pldm_fwup_pkg_creator.py
mynetz, line 1: Why is this copied? Can't we include via other means? I assume its overkill to consume the whole PLDM package. But copying something here that exists somewhere else in upstream feels somewhat wrong.

meta-canopy/scripts/gen-hpe-bios-tar
```diff
@@ -0,0 +1,236 @@
...
+		v) VERSION_OVERRIDE="$OPTARG" ;;
+		o) OUTPUT="$OPTARG" ;;
+		h) usage; exit 0 ;;
+		:) die "Option -$OPTARG requires an argument." ;;
+		\?) die "Unknown option: -$OPTARG" ;;
+	esac
+done
+
+[[ -n "$INPUT" ]] || die "Missing required argument: -i <input>"
+[[ -n "$MODEL" ]] || die "Missing required argument: -m <model>"
```
mynetz: I understand that this is copied from canopybmc/canopybmc#206 and your work - @pointbazaar - centers around other parts. Though I still don't like the requirement to provide the model even though it is encoded in the binary itself.
See: https://github.com/canopybmc/canopybmc/pull/206#discussion_r3473279269

 Judging by the code in the script I assume that this may require an additional translation step, but would allow for less human error potential. Although it seems that multiple servers have the same model short form. 
- U54: HPE ProLiant ML350/DL360/DL380 Gen11
- U59: HPE ProLiant DL560 Gen11
- U62: HPE ProLiant DL110 Gen11
- U63: HPE ProLiant DL320/ML110 Gen11
- A55: HPE ProLiant DL365/DL385 Gen11
- A56: HPE ProLiant DL325/DL345 Gen11
- A58: HPE ProLiant DL145 Gen11
- R11: HPE ProLiant RL300 Gen11

meta-canopy/scripts/pldm_fwup_pkg_creator.py
sinuscosinustan: The repo is not being included as submodule anyhwere or similar ways. I mean, I can add the full `pldm` / `libpldm` repo as submodule just for this one file. It is just easier to copy it and reference from where we've copied it.

meta-canopy/scripts/gen-hpe-bios-tar
sinuscosinustan: > Although it seems that multiple servers have the same model short form.

This is the exact reason why I didn't want to do it in the script until we got a confirmation from HPE on the mapping of SYSID and model name. I mean sure, we can just pin it to the already known ones for which we have UEFI files, U54, U63 and A55, but for the missing ones, even though they might be encoded in some way into the blob, we should have an official confirmation from HPE.
I'd also need to check the PLDM spec if multiple compatibles are allowed and if so, check for a possible limit of it.

mynetz: How about we at least print/list the hardware supported by the script in case the user misses the "model" parameter? I don't like to have this mandatory, but fine. It would probably only used by few people.

Though I don't get the point that we don't know the mapping. I added the list to the comment you replied to. That's obtained from the HPE website. They are not secret. The SYSID is referenced in some places.
We may not have the binaries for all to verify, but we know from 3 that this information is present. I don't expect that they change the binary format for one of the boards. iLO does probably also use the embedded information to validate a bios file.

sinuscosinustan: Added a known mapping to the bash script with those were we already got the BIOS files (SYSID to model)

mynetz:
It seems to me that the changes to the entity-manager files are pretty generic except for the `CompatibleHardware` name in the `FirmwareInfo`.
Couldn't we add this to all boards? I understand that the currently patched ones are the boards we have. If we want to play it safe I'd leave DL110, DL145 and DL560 out, but the other are at least present in OSFCI.

sinuscosinustan:
> Couldn't we add this to all boards?

Yes we can do that. I'll add it to the remaining ones.

sinuscosinustan (approved):
Approving Alex's work. Someone else has to review my commits before I press on merge

meta-canopy/scripts/gen-hpe-bios-pldm
```diff
@@ -0,0 +1,263 @@
...
+#
+# This will strip the (for us) arbritrary data and write the BIOS payload
+# into a PLDM firmware package that can be used to update the system
+# afterwards.
+#
+# The ROM version is auto-detected from the [ROM_INFO] block embedded
+# in the binary. Use `-v` to override it.
+#
+# Usage:
+#	gen-hpe-bios-tar -i <input> -m <model> [-v <version>] [-o <output>]
```
mynetz:
```suggestion
#	gen-hpe-bios-pldm‎ -i <input> [-m <model>] [-v <version>] [-o <output>]
```

meta-canopy/scripts/gen-hpe-bios-pldm
mynetz, line 1: Is this script intended to be run by the user manually? I don't see it reference somewhere else. I assume yes.

I needed to install these deps to run the called python script (on Ubuntu):
- python3-bitarray

I might had other dependencies already installed.

meta-canopy/scripts/gen-hpe-bios-pldm
```diff
@@ -0,0 +1,263 @@
...
+fi
+
+if [[ -n "$VERSION_OVERRIDE" ]]; then
+	VERSION="$VERSION_OVERRIDE"
+	echo "  Version:    $VERSION (from -v)"
+else
+	ROM_VER_STR="$(rom_info_get "$INPUT" ROM_VER_STR)"
+	[[ -n "$ROM_VER_STR" ]] || \
+		die "Could not find [ROM_INFO]/ROM_VER_STR in '$INPUT'. Use -v to specify version."
+	VERSION="$ROM_VER_STR"
```
mynetz:
```suggestion
	ROM_VER_X="$(rom_info_get "$INPUT" ROM_MAJOR_VER)"
	ROM_VER_Y="$(rom_info_get "$INPUT" ROM_MINOR_VER)"
	[ -n "$ROM_VER_X" || -n "$ROM_VER_Y" ]] || \
		die "Could not find [ROM_INFO]/ROM_VER_* in '$INPUT'. Use -v to specify version."
	VERSION="$ROM_VER_X.$ROM_VER_Y"
```
This is not mandatory but an observation. While `-v` can provide whatever string, the example is based on the pure version string. The one we extract from the binary is different.
While we don't enforce a version name - and the customer could do whatever - I think for consistency a fixe simple string would be nice.

mynetz (changes requested):
See comments for changes. I did also a manual test on the DL365.

Shows expected behavior. I can see the additional SPI flash for the host. I can update the BIOS and the drop-down is available.
I focused on the tooling side and ui. 

One thing I can observe is that the uploaded/updated image is shown as the "backup image" in the UI.

meta-canopy/scripts/gen-hpe-bios-pldm
sinuscosinustan: Yes. Specifically, the script must be run either by us or any user who has unencrypted BIOS binaries. I am not sure about who's getting those unencrypted binaries.

sinuscosinustan: Regarding the bitarray dependency: It's mentioned in the `Requirements` part of the script 😸

mynetz: Ah, you are right. Did miss that, thanks! Very nice.

mynetz (approved):

PR #211, opened 2026-07-01
ci: add dependabot

AtomicFS:
- replace update-openbmc.yml with dependabot
- this is much simpler and cleaner


mynetz:
Thanks @AtomicFS. I am sorry but I'll have to close that.

We debated usage of dependabot in canopybmc/canopybmc#177. You'll find that my proposal in https://github.com/canopybmc/canopybmc/pull/177#issuecomment-4378668158 pretty much covers what your PR does.

We plan to expand the rebase pipeline in the future to have smarter tooling to detect obsolete patches that we carry locallly to changes on upstream. Which would only be possible with dependabot, if it would also trigger another action that takes care of that. Which will then be as clean/unclean as the solution that currently triggers a script.

Also it couldn't produce the commit message in the way it was whished for. That, again, would be possible with using dependabot as a trigger for a workflow that could to that. But again, that wouldn't be simpler as the current solution.

PR #212, opened 2026-07-01
feat: add codeowners file

AtomicFS:
This is great feature for automatically ~~suggesting~~ **request** reviewers for PRs. Also defines responsibilities.

I had to kinda guess who would be in the file, so please feel free to correct my assumptions and propose changes.

I would also be comfortable with adding myself as responsible person for some files, but I do not want to just inject myself here. So I will await feedback.
```
/.github/workflows/  @sinuscosinustan @AtomicFS
```

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,8 @@
+# Documentation at:
+#   https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
+
+# Assign the entire repository (default ownership)
+*	@mynetz @pmaione @sylv-io @walterchris @sinuscosinustan
+
+/.github/workflows/	@sinuscosinustan @AtomicFS
+/.github/actions/	@sinuscosinustan @AtomicFS


mynetz:
How deep can the scoping go?

Right now it is probably fine to have a global list. Though I'd imagine once the official supported targets grow we might have primary maintainers based on that.
E.g. `meta-canopy/meta-hpe @sinuscosinustan`

There are some other parts of the project I see possible primary maintainers (eg. I could cover `doc`) but I wouldn't split that up now.

I am fine with you adding yourself to the GH workflow maintenance. The rest (global) for now can be covered by @sinuscosinustan. All the other current 9e contributers can help but are also assigned to other task that may interfere with time and priority.
They can be added as necessary to reviews. Right now thats:
- @walterchris
- @pmaione 
- @sylv-io 
- @mynetz (me)

AtomicFS:
I don't think that there is any limit. I have also included a link to docs.

AtomicFS:
@mynetz is this better?

sinuscosinustan (approved):

PR #213, opened 2026-07-01
ci: add dependabot

AtomicFS:
more lightweight version of #211

Only use dependabot to keep the GitHub actions up-to-date, do not touch the git submodules.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+---
+version: 2
+updates:
+  - package-ecosystem: 'github-actions'
+    directory: '/'
+    schedule:
+      interval: 'weekly'
+      time: '00:00'
+    groups:
+      all:
+        patterns:
+          - "*"


AtomicFS:
@mynetz review please?

mynetz:
@AtomicFS Sorry, yes. Thanks for the reminder.

Since we reverted the CODEOWNERS file, is there an option to have you assigned here by default to have a look? Doesn't mean you always have to do it. Just as the default.

.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
+---
+version: 2
+updates:
+  - package-ecosystem: 'github-actions'
+    directory: '/'
+    schedule:
+      interval: 'daily'
+      time: '00:00'
```
mynetz: I'd prefer weekly or monthly to reduce the noise.

.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
...
+    groups:
+      github:
+        update-types:
+          - 'minor'
+          - 'patch'
```
mynetz: Would it be possible to group more actions in one? I usually see 1-5 PRs for different actions in one scheduled cycle.

mynetz (changes requested):

AtomicFS:
@mynetz I don't think so. I think that only way to do automatic assignment is though codeowners file.

.github/dependabot.yml
AtomicFS: OK, then let's do weekly.

.github/dependabot.yml
AtomicFS: OK, I made it so that all updates will be grouped into single PR.

mynetz (approved):

sinuscosinustan (approved):

PR #214, opened 2026-07-01
ci: refactor build workflows

AtomicFS:
Here is my proposal for refactoring of the current CI. Next PR will be focused on adding release job, building on top of these changes.

The idea being:
- instead of copy-pasting workflow files for each target, let's have single template that uses [matrix strategy](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations) to run all target
- the list of targets is in `build-matrix.json` file (maybe not perfect, but we can change that)
- if there will be any changes needed for specific boards, we can use make the `.github/actions/build/action.yml` more parametric and use `include`:
```patch
 matrix:
   board: ${{ fromJson(needs.get-matrix.outputs.matrix) }}
+  include:
+    - board: canopy-qemu
+      additional_parameter: change-me
```

This approach might not be the most optimal, but it is IMHO large improvement over current way of doing things.


.github/workflows/build.yml
```diff
@@ -0,0 +1,84 @@
...
+      - main
+    paths-ignore:
+      - "doc/**"
+      - ".github/**"
+      - ".firmwareci/**"
+      - "README.md"
+  workflow_dispatch: {}
+
+jobs:
+  get-matrix:
```
walterchris, line 20: What's the upside of having a .json containing the matrix than directly having the matrix here in the .yml ? Would'nt that eliminate the parsing - and make this more readable + it's all in one place?

mynetz, line 20:
```suggestion
  get-matrix:
    name: Get list of build targets  
```

Not sure this is the best name, but I'd pefer somthing here. Since "get-matrix" looks a bit weird on the job listing. See: https://github.com/canopybmc/canopybmc/actions/runs/28526649605

.github/workflows/build-canopy.yml
```diff
@@ -0,0 +1,84 @@
+---
+name: canopy-qemu Build
```
mynetz:
```suggestion
name: canopy Build
```

Since this is now the generic build script remove the board/target name.

.github/workflows/build-canopy.yml
```diff
@@ -0,0 +1,84 @@
...
+
+on:
+  push:
+    branches:
+      - main
+    paths-ignore:
+      - "doc/**"
+      - ".github/**"
+      - ".firmwareci/**"
```
mynetz: Not sure why we did this in the original files. Excluding `.firmwareci` means that this is not run in case a PR is only adding tests. That seems wrong.

@walterchris original added this specifically to the list to trigger on. See: https://github.com/canopybmc/canopybmc/commit/862276468a18552136ad1e1235a5807bf8c9da6a

The changes from @sinuscosinustan then implicitly added this to the `paths-ignore`: https://github.com/canopybmc/canopybmc/commit/8419f856b94be3a80614c519018c9a4da321d6f9

There was some discussion about what to exclude/include. I am unsure. Whats the current opinion? Given my statement above, I'd say to remove the `.firmwareci` folder from the list.

.github/workflows/build-canopy.yml
```diff
@@ -0,0 +1,84 @@
...
+      - "README.md"
+  pull_request:
+    branches:
+      - main
+    paths-ignore:
+      - "doc/**"
+      - ".github/**"
+      - ".firmwareci/**"
```
mynetz: See: https://github.com/canopybmc/canopybmc/pull/214/changes#r3507331495

mynetz (changes requested):
I'd only make the workflow and action naming a blocking change. Rest was already present in the old files. Could stay in and be changed later. It was just something that popped into my eyes on the review.

.github/workflows/build-canopy.yml
walterchris: I think I agree - It means that when we are adding a tool, or DUT we still run the full CI - but I guess that's fine.

mynetz: Fine-tuning would probably be a headache. We have the workflows setup in a way that tie them to a specific hardware, but filtering this in the GH workflow would be too much.
Also I think in future there could be generic workflows for multiple DUTs?

This is more a feeling, but running some times on some HW unnecessarily is better than not running on test changes at all.

.github/workflows/build.yml
AtomicFS: Yeah, I think you are right. My main reason doing it this way was to make it human-readable.

But I am still not sure if GitHub can handle well multi-line JSON string. In case of separate JSON file, it can have newlines easily:
```json
[
  "canopy-qemu",
  "hpe-proliant-g11"
]
```

But in GitHub workflow, the typical usage is:
```yaml
run: |
  echo 'matrix=["canopy-qemu","hpe-proliant-g11"]' >> $GITHUB_OUTPUT
```

Maybe one way is to do this:
```yaml
run: |
  matrix=$(jq -cn '[
    "canopy-qemu",
    "hpe-proliant-g11"
  ]')
  echo "matrix=$matrix" >> "$GITHUB_OUTPUT"
```

AtomicFS: fair

AtomicFS:
Alright, what about this? I have reworked the matrix strategy, removed the JSON file, and also provided small example on how to use the `include` feature of the matrix strategy.

A demonstration with dummy job can be seen [here in this test job](https://github.com/canopybmc/canopybmc/actions/runs/28578913534?pr=214).

Also, the secrets like `FWCI_QEMU_WORKFLOW_NAME`, `FWCI_WORKFLOW_NAME` and `FWCI_MAIN_WORKFLOW_NAME` as they are not sensitive information, nor needed anymore.

mynetz:
@AtomicFS nice! Now I get what the `include` is for. Working around the not so generic build output..

AtomicFS:
OK, now finally ready for review, with FWCI working.

.github/workflows/build.yml
```diff
@@ -0,0 +1,105 @@
...
+      - name: Checkout
+        uses: actions/checkout@v7
+        with:
+          submodules: recursive
+          fetch-depth: 0
+
+      - name: Build
+        uses: ./.github/actions/build
+        with:
+          board: ${{ matrix.board }}
```
mynetz, line 80: You will need something to provide the signing key here as well. See:
https://github.com/canopybmc/canopybmc/blob/c75de6dea6b8309b936a676b9ef89ac96e2d0404/.github/workflows/build-hpe-proliant-g11.yml#L36-L37

AtomicFS: Is that always present? I do not see it for the QEMU target.

AtomicFS: Or maybe the question should be - can we have this there always for all targets?

mynetz: We do not care much for signing in the QEMU build. While right now the signing of the HPE board is special. All boards will have the options to be signed.

The HPE special case is that we have the HPE_SIGNING_KEY there while there generally is a SIGNING_KEY. I have a patch to unify this so that the HPE build will use "SIGNING_KEY" rather than "HPE_SIGNING_KEY".
That would then patch the line here to use the proper env. But doesn't matter now.
https://github.com/canopybmc/canopybmc/blob/c75de6dea6b8309b936a676b9ef89ac96e2d0404/.github/actions/build/action.yml#L32

You can assume for now that every build may have a signing key (one). That maybe change into a key reference when using a HSM.

AtomicFS: OK, then I will add
```
signing-key: ${{ secrets.HPE_9E_SIGNING_KEY }}
```
and you can patch it later.

mynetz: Question is if this is something to parameterize at some point via the include matrix. But I guess this can be done then. Yes. Go ahead this way.

mynetz (approved):
Okay, I'll approve it. The previous reviews are addressed. Also comments during the last review and this one are worked in. Fine by me.

mynetz:
I canceled the old CI job from a previous run to speed-up the testing of the latest state. Right now all looks good.

walterchris (approved):
Nice work.

AtomicFS:
@mynetz who does the merging? Also, merge via rebase, or with ugly merge commit?

PR #217, opened 2026-07-02
ci: add release job

### LINKED ISSUE #70: Canopy: Add, run and test release workflow
# Summary

Add a release workflow to the repository. The workflow shall allow to build specific boards for a given tag.

This shall be used to create a release (tagged version) once the feature work of the alpha milestone is completed.

## Tasks
- Create workflow with version/tag trigger
- Allow for two types: stable vesions/tags and releases

## References

- ...

AtomicFS:
Sorry, this might be a bit confusing, but I will do my best to explain. I had to work around some GitHub limitations, so that it is messier than I hoped.

The process of making releases is described in [doc/releases.md](https://github.com/canopybmc/canopybmc/blob/ci%2Frelease-job/doc/releases.md).

I had to create `release-prepare2.yml` workflow to work around GitHub limitations on workflows triggering other workflows (they do not want infinite loops, which makes sense).

Anyway, you manually trigger the `release-prepare` workflow, and you have to select branch (`main` by default) and provide version number. There is a [version number validation step](https://github.com/canopybmc/canopybmc/blob/6e4317e6e79e2dc00200e9cff08d7913a045e141/.github/workflows/release-prepare.yml#L25) with examples (I took the liberty of making a versioning scheme based on existing releases, I hope it is correct). If the provided version number is invalid, it will fail immediately and not carry on. This is to prevent accidental typos (they can still happen, but hopefully are less likely).

Once the `release-prepare` workflow finishes, it will create a pull request. After someone reviews it and merges it, everything from there on will be automatic process until a full release is complete.

I have also taken a liberty of creating `release-notes` directory, and populated it with existing release notes. There is a `README.md` explaining the file structure. In this directory is also `versions.json` file, which contains current and previous releases. This is so that the full changelog URL (`https://github.com/canopybmc/canopybmc/compare/2026.04-rc1...2026.04-rc2`) can be generated automatically, but I also use it heavily in the CI jobs.


.github/workflows/build.yml
```diff
@@ -104,3 +103,55 @@ jobs:
...
+      - name: Create source archives
+        run: |
+          zip -r "canopybmc-${{ steps.get_version.outputs.current_version }}.zip" "canopybmc-${{ steps.get_version.outputs.current_version }}"
+          tar -czvf "canopybmc-${{ steps.get_version.outputs.current_version }}.tar.gz" "canopybmc-${{ steps.get_version.outputs.current_version }}"
+
+      - name: Download all firmware images
+        uses: actions/download-artifact@v8
+        with:
+          path: binaries/
+          pattern: obmc-phosphor-image-*
```
sinuscosinustan, line 159: Needs to be fixed in the build pipeline: It should also export the signed tarballs (e.g. `obmc-phosphor-image-hpe-proliant-g11.static.mtd.tar` and `static.mtd.all.tar`). That way a user who uses our SRoT keys can update easily.

.github/workflows/build.yml
```diff
@@ -104,3 +103,55 @@ jobs:
...
+      - name: Debug
+        run: ls -a1lh -R binaries/
```
sinuscosinustan: leftover?

.github/workflows/release-prepare.yml
```diff
@@ -0,0 +1,87 @@
...
+      - name: Validate version
+        shell: bash
+        # Valid versions:
+        #   - 2025.01
+        #   - 2025.12
+        #   - 2025.07-rc
+        #   - 2025.07-rc1
+        #   - 2025.07-rc42
+        #   - 2025.07-lts
+        #   - 2025.07-lts-rc1
```
sinuscosinustan: I think LTS is not being tagged explicitly, right? @mynetz

.github/workflows/release-prepare.yml
```diff
@@ -0,0 +1,87 @@
...
+  release-prepare:
+    name: version_bump
+    runs-on: ubuntu-latest
+    steps:
+      - name: Validate version
+        shell: bash
+        # Valid versions:
+        #   - 2025.01
+        #   - 2025.12
+        #   - 2025.07-rc
```
sinuscosinustan, line 31: Patch releases like `2026.06.01` are valid too.

.github/workflows/release-prepare.yml
```diff
@@ -0,0 +1,87 @@
...
+        uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+          fetch-tags: true
+
+      - name: Update version
+        run: |
+          jq --arg new_version "${{ inputs.version }}" '. as $old | .current = $new_version | .previous = $old.current ' release-notes/versions.json > release-notes/versions2.json
+          mv release-notes/versions2.json release-notes/versions.json
+        # jq --arg new_version "${{ inputs.version }}" '. as $old | .next = $new_next | .current = $old.next | .previous = $old.current ' release-notes/versions.json
```
sinuscosinustan: leftover?

.github/workflows/build.yml
```diff
@@ -5,9 +5,8 @@ on:
   push:
     branches:
       - main
-    paths-ignore:
-      - "doc/**"
-      - "README.md"
```
sinuscosinustan, line 10: why was the paths-ignore being removed?

sinuscosinustan (changes requested):
The commit messages do not have any explanation about the goal, how it was solved and why certain removals were done. This is a requirement in our code-base to explain changes in the commit message too.

.github/workflows/release-prepare.yml
AtomicFS: Oh right, will fix that.

.github/workflows/build.yml
AtomicFS: Because it was messing with the release job. Basically, you still have the filter in the `pull_request` trigger, but every change on `main` should trigger this workflow. If that is a problem, we have to use something like [dorny/paths-filter](https://github.com/dorny/paths-filter) to filter out jobs only when not executed on `tag`.

sinuscosinustan: The explanation is fine for me, but please add it to the commit message with your explanation.

AtomicFS: Done

.github/workflows/build.yml
AtomicFS: Wait, what is `static.mtd.all.tar`? Is that produced for each board? Then this must be fixed in the `build-target` job, and not in here in `actions/download-artifact`.

It should be part of the uploaded tarball.

AtomicFS:
I have updated all commit messages to contain proper description. I hope this is good now.

.github/workflows/release-prepare.yml
mynetz: I think we didn't define anything. My assumption right now would be to not have it. Since I don't think we would have a 20XX.YY release and a 20XX.YY-lts variant. Since all releases are stable and "lts" is just a maintenance commitment. Put differently, it is not a version qualifier and shouldn't be in the release naming.

Though it reminds me that it has to be visible somewhere...

.github/workflows/release-prepare.yml
```diff
@@ -0,0 +1,87 @@
...
+  pull-requests: write # To create a pull request
+
+jobs:
+  release-prepare:
+    name: version_bump
+    runs-on: ubuntu-latest
+    steps:
+      - name: Validate version
+        shell: bash
+        # Valid versions:
```
mynetz, line 27: @AtomicFS I am sorry - we didn't talk about that before. The version names should carry information on the stability of the release. With potential for:
- `alphaX` (maybe)
- `betaX`
- `rcX`
- \<release\>

Though I deem alpha unlikely since this is somewhat similar to a build after a merge (or nightly in case this becomes a thing)

.github/workflows/release-prepare.yml
```diff
@@ -0,0 +1,87 @@
...
+        #   - 2025.07-lts-rc1
+        # Invalid:
+        #   - 2025.13     (invalid month)
+        #   - 2025.00     (invalid month)
+        #   - 25.07       (year not 4 digits)
+        #   - 2025.7      (month not 2 digits)
+        #   - 2025.07-rcx (suffix after rc must be digits)
+        run: |
+          VERSION="${{ inputs.version }}"
+          REGEX='^[0-9]{4}\.(?:0[1-9]|1[0-2])(?:\.[0-9]+)?(?:-lts)?(?:-rc(?:[0-9]+)?)?$'
```
mynetz: I don't think we should stack the qualifiers.
<year>.<month>[.patchlevel][-qualifier]

See https://github.com/canopybmc/canopybmc/pull/217/changes#r3534224639 for qualifier discussion.

.github/workflows/release-tag-create.yml
mynetz, line 1: I try to understand why this is needed as a separate job. Why can't the first one tag? I assume that is because it would do this when the action runs and it would be better to do this after the PR is merged since this is the point where we decided the release is fine this way.

.github/workflows/release-prepare2.yml
```diff
@@ -0,0 +1,67 @@
...
+      - closed
+
+permissions:
+  contents: write # To push new branch
+  actions: write # To trigger another workflow
+
+jobs:
+  release-prepare2:
+    runs-on: ubuntu-latest
+    if: startsWith(github.event.pull_request.title, 'Release:') && github.event.pull_request.merged == true
```
mynetz: This probably means that the "Release:" naming in PRs is restricted to just the release job. I wonder if this is a footgun that is lying around...
Maybe, just in case we could also add and filter for "(release-prepare)" at the end of the job name to scope it to the previous action.

doc/releases.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Description
+
+The release CI is a bit complicated, to accommodate the project specific quirks.
+
+There are following workflows:
+- `release-prepare.yml`
+    - this workflow can be triggered manually, it will open a pull request with all changes needed for a new release
+- `release-prepare2.yml`
+    - this workflow is triggered automatically when the pull request made by `release-prepare.yml` is merged
```
mynetz, line 24: Yes. Though maybe add a hint that it in fact depends on the name of the previously created PR.

release-notes/2026.04-rc1.md
mynetz, line 1: Thanks for including the existing RNs! 👍

release-notes/README.md
mynetz, line 1: Does the template and system in place right now allow for a future workflow in which part of the RNs come from properly tagged Issues/PRs? Or would that need some rework in the release process?

I am fine, if this is not a 100% set right now. Just want to know.

I assume this would be done similar to the `SED_REPLACE_` magic for the history of the release.

mynetz (changes requested):
The process is now to have a phase one of the release that allows to review the release notes and other release related generated things, and then after merging trigger to have automagic release process. The binaries are produced last.

What I don't like with this is that the crucial part of the release - the binary artifacts - are done afterwards and if they fail in the CI - for whatever reason - they are still in the release. (at least this is how I read it right now)

With all you have evaluated, could you imagine that it would be possible to have a release trigger that prepares the release (eg. generates the RNs) then builds this and have the release prepared in some way. Then have all the results in the CI.
Merging/accepting will then puplish the release.
Tagging would then just be done after the fact with the SHA of the branch at the time of release.
If this is not possible we could have intermediate release tags maybe.

.github/workflows/release-tag-create.yml
walterchris, line 1: Can we rename this? I mean it's essentially triggering a release (and not preparing it) - I don't like the 2 amended on the filename and the job itself.

.github/workflows/build.yml
```diff
@@ -104,3 +103,53 @@ jobs:
...
           GITHUB_INSTALLATION_ID: ${{ secrets.FWCI_GITHUB_INSTALLATION_ID }}
+
+  release:
+    runs-on: ubuntu-latest
+    if: github.ref_type == 'tag'
+    needs:
+      - build-target # needed for artifacts
+      - firmwareci # we want to only release on functional and tested codebase
+    steps:
+      - name: Checkout
```
walterchris, line 118: Is there a benefit to check the repo out with depth 0 before you check it out again? is it to fail fast if we don't find the tag?

.github/workflows/release-tag-create.yml
walterchris: Hups - I guess we need to rename the other.

.github/workflows/release-prepare.yml
walterchris, line 1: So this needs to be renamed to release-trigger.yml

.github/workflows/release-prepare.yml
AtomicFS: Makes sense

.github/workflows/release-tag-create.yml
AtomicFS: Talked about in a meeting

.github/workflows/release-prepare2.yml
AtomicFS: I will add filter to also check the author.

.github/workflows/release-tag-create.yml
AtomicFS: I can rename it.

.github/workflows/release-prepare.yml
AtomicFS: The other way around ;)

.github/workflows/build.yml
AtomicFS: Yes, the first time we need it to access the `release-notes/versions.json` file. The second time it checkout into a specific sub-directory `canopybmc-${{ steps.get_version.outputs.current_version }}` (so that when the ZIP / TAR.GZ archive is extracted, the directory has version in the name).

And even if we did not want the version to be in the dirname, I would still do that so that we can easily create a clean archive without and runtime trash that might congregate in the `./`.

release-notes/README.md
AtomicFS: Yes, it is possible to automatically update release notes with data from GitHub issue tracker and PR list.

Steps to automatically enhance / auto-complete / create / alter the release notes, such steps should go into `release-prepare.yml`, before or after the `Update release notes` step.

I hope that answers the question :)

AtomicFS:
I have made changes according to your feedback from reviews and from the meeting today. And some small bug-fixes.

On to of that, I have also fixed the links in release notes. Before it was just GitHub reference like `#152` which only when viewed in GitHub would appear as URL link. So I have changed them to proper Markdown URL links. While doing that, the full changelog URL is not proper Markdown link.

I have also re-introduced the `paths-ignore`, since after reading the docs I think it should be fine. The push to `tag` triggers the workflow, the `paths-ignore` is not evaluated. So should be fine.

Added note about this into `doc/releases.md` along with some other topics we discussed during our meeting that were unclear.

.github/workflows/build.yml
sinuscosinustan, line 1: You've re-added paths-ignore but didn't change the commit message ;)

sinuscosinustan (changes requested):

.github/workflows/build.yml
AtomicFS: Oh yeah, LOL

.github/workflows/build.yml
sinuscosinustan: `static.mtd.all.tar` is produced for all boards by default (to be precise: for all that use the static mtd layout, which is the case for us.) It can be used to re-flash the BMC (incl. the rwfs), but it doesn't reflash the GXP bootloader.

So yeah, that must be fixed in the build-target.

mynetz: The previous build jobs did only focus on the binary result(s) that can be flashed. I think there was no example to draw from and we didn't specify it until now.

Right now we filter by `<board>.*static.mtd` for the uploads.
https://github.com/canopybmc/canopybmc/blob/173e4b3fb9d47504d62f900ca9982b49247cfbfc/.github/actions/build/action.yml#L38-L42

You want this to be changed to `<board>.*static.mtd*` to select `mtd`, `mtd.tar` and `mtd.all.tar`? Do we always need this? Seems to be a lot of data we accumulate over time there.

mynetz (commented):
PR okay. I would approve it.

There is one unresolved discussion to archive/export more artifacts.
https://github.com/canopybmc/canopybmc/pull/217#discussion_r3530864446

I would like to see this resolved - either by including the binaries or have a decision not to. I think we'll include them since they are for the normal upgrade path. I would just like to see a final statement from @sinuscosinustan here.

.github/workflows/build.yml
AtomicFS: Then I propose that in release, we do not upload individual files, but compressed ZIP files (one for each target).

So instead of this:
```
- canopybmc-202x.xx.tar.gz
- canopybmc-202x.xx.zip
- obmc-phosphor-image-canopy-qemu.static.mtd
- obmc-phosphor-image-hpe-proliant-g11.GXP2loader-t26x-sgn00.static.mtd
- obmc-phosphor-image-hpe-proliant-g11.GXP2loader-t277-t280-t285-sgn00.static.mtd
- obmc-phosphor-image-hpe-proliant-g11.GXP2loader-t282-t288-sgn00.static.mtd
- obmc-phosphor-image-hpe-proliant-g11.static.mtd
- Source code (zip)
- Source code (tar.gz) 
```

We will have this:
```
- canopybmc-202x.xx.tar.gz
- canopybmc-202x.xx.zip
- obmc-phosphor-image-canopy-qemu.zip
- obmc-phosphor-image-hpe-proliant-g11.zip
- sbom-canopy-qemu.zip
- sbom-hpe-proliant-g11.zip
- Source code (zip)
- Source code (tar.gz)
```

This makes more sense, and also then we do not have to be afraid of filename collisions. Then we just release whatever the `build-target` step produces.

As for the `build-target` step (or `.github/actions/build`), I don't know how to alter it to produce the files you want. Please feel free to make a commit on top of this branch.

AtomicFS: Talking about altering the `build-target` step (or `.github/actions/build`) ... do we want separate firmware images and SBOM? I think they should be combined into single ZIP, so that we have only:
```
- canopybmc-202x.xx.tar.gz
- canopybmc-202x.xx.zip
- obmc-phosphor-image-canopy-qemu.zip
- obmc-phosphor-image-hpe-proliant-g11.zip
- Source code (zip)
- Source code (tar.gz)
```

We can do
```patch
-- name: Upload firmware image
+- name: Upload firmware image with SBOM
   uses: actions/upload-artifact@v7
   with:
     name: obmc-phosphor-image-${{ inputs.board }}
-    path: build/${{ inputs.board }}/tmp/deploy/images/${{ inputs.board }}/obmc-phosphor-image-${{ inputs.board }}.*static.mtd
+    path: |
+      build/${{ inputs.board }}/tmp/deploy/images/${{ inputs.board }}/obmc-phosphor-image-${{ inputs.board }}.*static.mtd
+      build/${{ inputs.board }}/tmp/deploy/images/${{ inputs.board }}/obmc-phosphor-image-${{ inputs.board }}.spdx.json
-
-- name: Upload SBOM
-  uses: actions/upload-artifact@v7
-  with:
-    name: sbom-${{ inputs.board }}
-    path: build/${{ inputs.board }}/tmp/deploy/images/${{ inputs.board }}/obmc-phosphor-image-${{ inputs.board }}.spdx.json
```

.github/workflows/release-prepare.yml
```diff
@@ -0,0 +1,95 @@
...
+        uses: peter-evans/create-pull-request@v8
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          branch: release/${{ inputs.version }}
+          commit-message: "chore: bump version to ${{ inputs.version }}"
+          title: "Release: ${{ inputs.version }} Pull Request"
+          body: "Prepare for next release"
+          # TODO: Not sure who should be responsible
+          assignees: "AtomicFS"
+          reviewers: "AtomicFS"
```
AtomicFS: Also, please tell me if I should assign someone here.

mynetz: Can we use `github.actor` (or `github.triggering_actor`)
https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#github-context

The one triggering the `release-prepare` is the assignee and would be the primary reviewer and can add others.

AtomicFS: That would be difficult to do, because then we have to somehow pass this information though multiple workflows, or query API to get `github.actor` from the last `release-prepare.yml` workflow run.

These fields are optional, so we could also omit them ;)

mynetz: The github.actor should stay the same for the full workflow. At least this is how I read the doc. It would be perfect since it reflects the user that did the action and we don't have to have someone static here.

If that doesn't work. Leave it empty.

.github/workflows/build.yml
mynetz: I am in favor of having a single file (zip) per platform. Inlcuding the SBOM.

Note that the FwCI job will probably need some modification to extract the proper files from the ZIP then.

.github/workflows/release-prepare.yml
AtomicFS: Except we have 3 different workflows xD

`release-prepare.yml` -> `release-tag-create.yml` -> `build.yml`

So I guess lets leave it empty then.

mynetz: Yes, but your question was on the `release-prepare.yml` and the author its job sets for the created PR. There it should be constant.

AtomicFS: Oh wait, I was being silly :D it is in one workflow in this case, sorry :D

AtomicFS: Works

.github/workflows/build.yml
AtomicFS: OK, I have made the necessary changes to have single ZIP file per target. Now we just need to update the file of files. @sinuscosinustan can you either do it, or provide the `path:` filters for me to amend?

AtomicFS: Fixed

sinuscosinustan (approved):
LGTM! Thank you for tackling all of this!

@mynetz would you mind approving this too?

AtomicFS:
Happy to help :)

mynetz (approved):
I am happy now :)

PR #226, opened 2026-07-06
meta-hpe: em: remove static asset tag

sinuscosinustan:
This was added by HPE before for unknown reason and shouldn't be there.

### DIFF
--- a/.firmwareci/workflows/hpe-dl320-g11-ci/tests/inventory-system-identity.yaml
+++ b/.firmwareci/workflows/hpe-dl320-g11-ci/tests/inventory-system-identity.yaml
@@ -3,7 +3,7 @@ description: >-
   Verify chassis and system identity fields are populated correctly via Redfish.
   Covers entity-manager VPD patches, devicetree-vpd-parser, and SMBIOS system
   identity. Checks both Chassis and System resources for manufacturer, model,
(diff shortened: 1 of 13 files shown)


mynetz:
Out of curiosity, how is that handled on other platforms? Isn't the asset tag something the vendor defines in production (at least this seems to be the case for Dell). Would we need to decode that somewhere? Or is that a field the user can set at some point?

sinuscosinustan:
Usually the asset tag is stored on the FRU EEPROM. However, we don't have access to that. Fortunately `phosphor-inventory-manager` creates an asset tag property on `/system` by default. I've adapted the firmware ci test to patch the asset tag before which ultimately writes it to PIM.

sinuscosinustan:
Marking as draft. Need to check some stuff why it HTTP 500s now...

sinuscosinustan:
Well... I thought it'd be easy, but it isn't...

`bmcweb` queries the item that holds `Item.System`, but this is held by `entity-manager`. Now, I could adapt `bmcweb` to use `/system` for the asset tag, but it's unlikely that this gets into upstream... I'd say remove `AssetTag` completely for now on this platform as we can't change it easily. Maybe there's something in the EEPROM, but I wouldn't be too confident on that.

PR #239, opened 2026-07-23
HPE ProLiant Gen11: NVRAM/BIOS settings reset support

sinuscosinustan:
This PR consists of multiple commits, but in summary they address:

- Backporting the pending `cmos-reset` topic to `platform-init` and `bmcweb`, so that we can actually reset the BIOS NVRAM.
- Bringing up `platform-init` on the HPE GXP platforms with NVRAM reset support (erases `/dev/mtd/host-reserved`, which holds both the NVRAM and the BIOS EVs, and also clears the EV storage in `/var/lib/chif/evs.dat`)
- Adding the memory-mapped host NVRAM partition (`host-reserved`) to the GXP device tree (needed before any of the NVRAM reset stuff works)
- FWCI: Reset the NVRAM before the first boot for a clean state on certain tests (doing it for every test would increase the CI time too much)

PR #241, opened 2026-07-24
FWCI Cleanup for HPE ProLiant DL320 Gen11

sinuscosinustan:
These commits were cherry-picked from #239 to split the amount of changes a bit more up (and to not mix-up the goal of each PR).

- The SMBIOS timeout must be increased to 15 minutes as the 8 minutes aren't enough after certain tests. 15 minutes should be plenty enough of time.
- Remove duplicate tasks to the pre-stage (DRY)
- Power off the system via Redfish before cutting the power, as the system would recover to the last power state otherwise.


walterchris (approved):
LGTM

PR #247, opened 2026-07-29
meta-canopy: scripts: fix automatic version detection

haeckle:
just a small fix to make the automatic host firmware detection work in the PLDM generation script


sinuscosinustan (approved):
I seem to have forgotten to open a PR for the fix lol. Approved, thanks!

PR #248, opened 2026-07-29
HPE ProLiant Gen11: Fix PlatDef Multi-bundle parsing

sinuscosinustan:
This PR consists of multiple commits.

The APML blob on the host SPI can contain several PlatDef bundles, one per platform, and we unconditionally took the first one. On multi-platform images that meant parsing another platform's PlatDef and ending up with a wrong I2C segment map (a DL320 Gen11 may have used the ML110 Gen11 bundle, depending on the order on the SPI).

We now walk the bundle chain and match the platform ID against the `ProductId` of the inventory object.

### DIFF
--- a/meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/platdef_extract.cpp
+++ b/meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/platdef_extract.cpp
@@ -9,29 +9,42 @@
 #include "platdef_extract.hpp"
 
 #include "uefi_fv.hpp"
-
-#include <phosphor-logging/lg2.hpp>
+#include "utils.hpp"
 
 #include <zlib.h>
 
+#include <phosphor-logging/lg2.hpp>
+#include <xyz/openbmc_project/Inventory/Item/Board/Motherboard/common.hpp>
+
+#include <algorithm>
+#include <cstddef>
 #include <cstring>
 #include <filesystem>
 #include <fstream>
+#include <limits>
+#include <optional>
+#include <ranges>
+#include <span>
 #include <string>
 #include <string_view>
 
 namespace chif
 {
 
+static constexpr std::string_view bundleSignature = "$PlatdefBundle1$";
+
+using Motherboard = sdbusplus::common::xyz::openbmc_project::inventory::item::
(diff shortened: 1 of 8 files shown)


meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/platdef_extract.cpp
```diff
@@ -171,40 +183,97 @@ static std::string findMtdByLabel(const std::string& label)
...
     {
-        lg2::error("PlatDef: invalid bundle signature");
-        return {};
+        const auto* bundle =
+            reinterpret_cast<const PlatDefBundleHeader*>(raw.data() + offset);
+
+        size_t headerBytes = static_cast<size_t>(bundle->headerLength) * 16;
+        size_t totalSize = bundle->totalSize;
+
+        if (headerBytes < sizeof(PlatDefBundleHeader) +
```
pmaione: This kind of conditional is very hard to read/debug. Maybe rewrite this to improve readability?
```
if ((condition) || 
    (condition) || ...
    (condition)) {
    ...
}

Comment (pmaione) [reply meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/platdef_extract.cpp]: A better overall solution to this kind of formatting would be including a clang-format file.

**Review by pmaione — APPROVED** [review]
LGTM!
With a minor formatting issue.

File: meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/platdef_extract.cpp (hunk 1 above)
Comment (sinuscosinustan) [reply meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/platdef_extract.cpp]: Interesting :thinking: I ran clang-format with the OpenBMC config against it. I'll improve it manually

Comment (sinuscosinustan) [reply meta-canopy/meta-hpe/meta-common/recipes-hpe/gxp-chif-service/gxp-chif-service/src/platdef_extract.cpp]: Yeah, running `clang-format` auto-corrects it to:
```patch
+        if ((headerBytes < bundleSize) || (totalSize <= headerBytes) ||
+            (offset + totalSize > raw.size()))
```

I've moved the inline `bundleSize` math into a separate variable for readability.

PR #252, opened 2026-08-05
HPE ProLiant Gen11: Firmware CI fixes

sinuscosinustan:
This PR fixes observed issues in our Firmware CI environment.
- Fix the DIMM location in the memory inventory test (physically moved as it was not populated correctly)
- Move the NVRAM restore to the post stage to prevent altering from other tests
- Fix the Reboot stability test by waiting for SMBIOS transfer and restoring the EVs from FWCI storage too.


pmaione (approved):
LGTM. Great work!

Checked the FWCI logs and everything runs OK!

Issue #92, opened 2026-03-12
HPE Proliant Gen11: UEFI reports CHIF response errors


sinuscosinustan:
# Issue Summary

The UEFI reports several `CHIF response invalid` errors during boot-up via the SOL (only a fracture of the logs):

```
ERROR: C80000002:V801000A2 I0 85CC8B8F-F2B6-11E1-B12E-3C4A92B21913 637F2510
DownloadDemandedTypePlatDef: CHIF response invalid! Status = Device Error
GetMegacellI2CInfo: DownloadDemandedTypePlatDef failed! Status = Device Error
HpIicWriteRead: CHIF response invalid.  Size 12.  Cmd 32882.  Sid 0
HpIicWriteRead: CHIF response invalid.  Size 12.  Cmd 32882.  Sid 0
HpIicWriteRead: CHIF response invalid.  Size 12.  Cmd 32882.  Sid 0
ReadI2CMegacellVerify: I2C Read Chif Transaction Failed,Bus=15,Addr=C0,Reg=0,Error = Device Error
HpIicWriteRead: CHIF response invalid.  Size 12.  Cmd 32882.  Sid 0
HpIicWriteRead: CHIF response invalid.  Size 12.  Cmd 32882.  Sid 0
HpIicWriteRead: CHIF response invalid.  Size 12.  Cmd 32882.  Sid 0
ReadI2CMegacellVerify: I2C Read Chif Transaction Failed,Bus=15,Addr=C0,Reg=3A,Error = Device Error
```

Apparently we silently drop lots of messages or have no implementation for them.

## Expected behavior

No CHIF errors from UEFI.

## Current behavior

UEFI reports errors on bootup via SOL which _may_ cause issues in the future.

## Workaround

No

## Build environment

```
OS: Ubuntu 24.04 LTS
Python: 3.12
GCC: 13.3.0-6ubuntu2~24.04.1
Selected build target: hpe-proliant-g11
```

PR #253, opened 2026-08-07
FirmwareCI: dl320g11: add host firmware update test

haeckle:
Testing the update feature added in #210 

- should we also test a downgrade?
- golden image currently on 2.72
- without the "prepare NVRAM" stage it currently fails to boot the OS (reset loop)

Latest run: https://app.firmware-ci.com/canopy/jobs/01KZXCCXNS2J7TRM7694G33SW4?projectID=01KKGTVRZ3XK1ZDSAETKB1624Y

Closes #242

### DIFF
--- a/.firmwareci/duts/dut-hpe-dl320/pre.yaml
+++ b/.firmwareci/duts/dut-hpe-dl320/pre.yaml
@@ -16,6 +16,14 @@ pre-stage:
       args: ["[[input.firmware]]"]
 
   - cmd: dutctl
+    name: Emulate BIOS Flash with golden image
+    parameters:
+      agent: "[[attributes.Agent]]"
+      device: "[[attributes.Device]]"
+      command: emulate-host
+      args: ["[[storage.dl320g11_bios_update]]/[[storage.dl320g11_bios_update.golden_image]]"]
+
+  - cmd: dutctl
     name: Power on DUT
     parameters:
       agent: "[[attributes.Agent]]"
--- a/.firmwareci/duts/dut-hpe-dl320/pre.yaml
+++ b/.firmwareci/duts/dut-hpe-dl320/pre.yaml
@@ -70,3 +70,23 @@ pre-stage:
       timeout: 2m
     parameters:
       command: "journalctl -t phosphor-bmc-state-manager -f | grep -m 1 BMC_READY"
+
+  - cmd: copy
+    name: Copy original NVRAM to DUT
+    transport: *bmc_ssh
+    parameters:
+      source: "[[storage.dl320g11_nvram]]"
+      destination: /tmp
+      recursive: true
+
+  - cmd: shell
+    name: Copy EVs to CHIF path
+    transport: *bmc_ssh
+    parameters:
+      command: "cp /tmp/[[storage.dl320g11_nvram.evs]] /var/lib/chif/evs.dat"
+
+  - cmd: shell
+    name: Restart CHIF service
+    transport: *bmc_ssh
+    parameters:
+      command: "systemctl restart xyz.openbmc_project.GxpChif.service"
--- a/.firmwareci/duts/dut-hpe-dl320/pre.yaml
+++ b/.firmwareci/duts/dut-hpe-dl320/pre.yaml
@@ -58,12 +58,15 @@ pre-stage:
     parameters:
       host: "[[attributes.BMC]]"
 
-  - cmd: cmd
-    name: Wait for BMC services
-    transport: &local
-      proto: local
+  - cmd: shell
+    name: Wait for BMC ready
+    transport: &bmc_ssh
+      proto: ssh
+      options:
+        host: "[[attributes.BMC]]"
+        user: root
+        password: "[[attributes.BMCPassword]]"
     options:
       timeout: 2m
     parameters:
-      executable: sleep
-      args: ["45"]
+      command: "journalctl -t phosphor-bmc-state-manager -f | grep -m 1 BMC_READY"
--- a/.firmwareci/workflows/hpe-dl320-g11-ci/tests/host-firmware-update.yaml
+++ b/.firmwareci/workflows/hpe-dl320-g11-ci/tests/host-firmware-update.yaml
@@ -24,28 +24,6 @@ variables:
   UpdateTriggerOutput: null
 
 stages:
-  - name: Prepare NVRAM # TODO this shouldn't be needed, investigate
-    steps:
-      - cmd: copy
-        name: Copy original NVRAM to DUT
-        transport: *bmc_ssh
-        parameters:
-          source: "[[storage.dl320g11_nvram]]"
-          destination: /tmp
-          recursive: true
-
-      - cmd: shell
-        name: Copy EVs to CHIF path
-        transport: *bmc_ssh
-        parameters:
-          command: "cp /tmp/[[storage.dl320g11_nvram.evs]] /var/lib/chif/evs.dat"
-
-      - cmd: shell
-        name: Restart CHIF service
-        transport: *bmc_ssh
-        parameters:
-          command: "systemctl restart xyz.openbmc_project.GxpChif.service"
-
   - name: Host on, OS Booting
     steps:
       - cmd: shell
--- a/.firmwareci/workflows/hpe-dl320-g11-ci/tests/host-firmware-update.yaml
+++ b/.firmwareci/workflows/hpe-dl320-g11-ci/tests/host-firmware-update.yaml
@@ -1,0 +1,228 @@
+name: "Host Firmware: Update"
+description: >-
+  Update the host firmware (aka BIOS) via Redfish and verify it is booting the
+  new one.
+
+  For HPE some magic firmware preparation is needed beforehand to get a correct
+  PLDM package.
+
+  If flashing takes to long the Redfish task is running in a timeout and therefore
+  gets cancelled. The flash operation is however uneffected by this, so we don't
+  error out. "Cancelled" = timeout, "Exception" = error in the update process.
+  Possible errors are captured in the following step through systemd-journal.
+
+defaults:
+  transport: &bmc_ssh
+    proto: ssh
+    options:
+      host: "[[attributes.BMC]]"
+      user: root
+      password: "[[attributes.BMCPassword]]"
+
+variables:
+  UpdateTarget: null
+  UpdateTriggerOutput: null
+
+stages:
+  - name: Prepare NVRAM # TODO this shouldn't be needed, investigate
+    steps:
+      - cmd: copy
+        name: Copy original NVRAM to DUT
+        transport: *bmc_ssh
+        parameters:
+          source: "[[storage.dl320g11_nvram]]"
+          destination: /tmp
+          recursive: true
+
+      - cmd: shell
+        name: Copy EVs to CHIF path
+        transport: *bmc_ssh
+        parameters:
+          command: "cp /tmp/[[storage.dl320g11_nvram.evs]] /var/lib/chif/evs.dat"
+
+      - cmd: shell
+        name: Restart CHIF service
+        transport: *bmc_ssh
+        parameters:
+          command: "systemctl restart xyz.openbmc_project.GxpChif.service"
+
+  - name: Host on, OS Booting
+    steps:
+      - cmd: shell
+        name: Power on host
+        transport:
+          proto: local
+        parameters:
+          command: >-
+            curl -sk -u root:[[attributes.BMCPassword]]
+            -X POST -H 'Content-Type: application/json'
+            -d '{"ResetType":"On"}'
+            -o /dev/null -w '%{http_code}'
+            https://[[attributes.BMC]]/redfish/v1/Systems/system/Actions/ComputerSystem.Reset
+          expect:
+            - regex: "^(200|204)$"
+
+      - cmd: shell
+        name: Wait for SMBIOS transfer (host POST complete)
+        transport: *bmc_ssh
+        options:
+          timeout: 10m
+        parameters:
+          command: >-
+            while [ ! -f /var/lib/smbios/smbios2 ] ||
+                  [ $(stat -c %s /var/lib/smbios/smbios2 2>/dev/null || echo 0) -le 1000 ];
+            do sleep 1; done;
+            echo "SMBIOS ready: $(stat -c %s /var/lib/smbios/smbios2) bytes"
+          expect:
+            - regex: "SMBIOS ready"
+
+      - cmd: ping
+        name: Wait for host OS
+        options:
+          timeout: 3m
+        parameters:
+          host: "[[attributes.Host]]"
+
+      - cmd: redfish
+        name: Current BiosVersion
+        options:
+          timeout: 10s
+        transport: &local
+          proto: local
+        parameters:
+          host: "[[attributes.BMC]]"
+          auth:
+            user: root
+            password: "[[attributes.BMCPassword]]"
+          validate: inventory
+          inventory:
+            rules:
+              - path: /redfish/v1/Systems/system
+                prop: BiosVersion
+                op: matches
+                value: \d+\.\d+
+
+  - name: Firmware update to 2.92
+    steps:
+      - cmd: shell
+        name: Get update endpoint
+        options:
+          timeout: 30s
+        transport: *local
+        parameters:
+          command: |
+            curl -sk --user root:[[attributes.BMCPassword]] \
+              https://[[attributes.BMC]]/redfish/v1/UpdateService/FirmwareInventory \
+              | grep -Eo UEFIPrimary_[0-9]+
+          expect:
+            - regex: 'UEFIPrimary_\d+'
+          variable: UpdateTarget
+
+      - cmd: redfish
+        name: Is updateable?
+        options:
+          timeout: 30s
+        transport: *local
+        parameters:
+          host: "[[attributes.BMC]]"
+          auth:
+            user: root
+            password: "[[attributes.BMCPassword]]"
+          validate: inventory
+          inventory:
+            rules:
+              - path: /redfish/v1/UpdateService/FirmwareInventory/{{ UpdateTarget }}
+                prop: Updateable
+                op: eq
+                value: "true"
+
+      - cmd: shell
+        name: Trigger update
+        options:
+          timeout: 5m
+        transport: *local
+        parameters:
+          command: >-
+            curl -sk --user root:[[attributes.BMCPassword]]
+            -X POST -H Content-Type:multipart/form-data
+            -F 'UpdateParameters={"Targets":["/redfish/v1/UpdateService/FirmwareInventory/{{ UpdateTarget }}"],"@Redfish.OperationApplyTime":"Immediate"};type=application/json'
+            -F 'UpdateFile=@[[storage.dl320g11_bios_update]]/[[storage.dl320g11_bios_update.pldm]];type=application/octet-stream'
+            https://[[attributes.BMC]]/redfish/v1/UpdateService/update-multipart
+          expect:
+            - regex: TaskStarted
+          variable: UpdateTriggerOutput
+
+      - cmd: shell
+        name: Wait for update finish (Redfish)
+        # fails when update is too slow
+        transport: *local
+        options:
+          timeout: 10m
+          on-error: continue
+        parameters:
+          command: |
+            Task=$(echo {{ UpdateTriggerOutput }} | grep -Eo "/redfish/v1/TaskService/Tasks/[0-9]+" );
+            TaskState="Running";
+            while [ "$TaskState" = "Running" ]; do
+              sleep 2;
+              TaskState=$(curl -sk --user root:[[attributes.BMCPassword]] https://[[attributes.BMC]]${Task} | grep TaskState | sed -E 's/.*"(.*)",$/\1/');
+            done;
+            echo "$TaskState"
+          expect:
+            - regex: Completed
+
+      - cmd: shell
+        name: Wait for update finish (systemd-journal)
+        options:
+          timeout: 5m
+        transport: *bmc_ssh
+        parameters:
+          command: |
+            journalctl -n 5 -t phosphor-bios-software-update -t power-control -f \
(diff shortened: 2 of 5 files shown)


.firmwareci/workflows/hpe-dl320-g11-main/tests/host-firmware-update.yaml
```diff
@@ -0,0 +1,386 @@
...
+        name: Copy EVs to CHIF path
+        transport: *bmc_ssh
+        parameters:
+          command: "cp /tmp/[[storage.dl320g11_nvram.evs]] /var/lib/chif/evs.dat"
+
+      - cmd: shell
+        name: Restart CHIF service
+        transport: *bmc_ssh
+        parameters:
+          command: "systemctl restart xyz.openbmc_project.GxpChif.service"
```
sinuscosinustan, line 44: The NVRAM copy might not be needed, but the EVs are. The reason for it is due to how we do our testing it can happen that the NVRAM gets so heavily unusable that you cannot boot the system anymore without manual intervention.

.firmwareci/workflows/hpe-dl320-g11-main/tests/host-firmware-update.yaml
```diff
@@ -0,0 +1,386 @@
...
+        transport:
+          proto: local
+        parameters:
+          command: |
+            NEWTARGET=$(curl -sk --user root:[[attributes.BMCPassword]] \
+            https://[[attributes.BMC]]/redfish/v1/UpdateService/FirmwareInventory \
+            | grep -Eo UEFIPrimary_[0-9]+)
+            if [ {{ UpdateTarget }} = $NEWTARGET ]; then
+              echo $NEWTARGET;
+            fi;
```
sinuscosinustan, line 196: Make sure to quote both strings

.firmwareci/workflows/hpe-dl320-g11-main/tests/host-firmware-update.yaml
```diff
@@ -0,0 +1,386 @@
...
+        transport:
+          proto: local
+        options:
+          timeout: 10s
+        parameters:
+          command: |
+            BIOSVER="null";
+            while [ "$BIOSVER" = "null" ]; do
+              sleep 2;
+              BIOSVER=$(curl -sk --user root:[[attributes.BMCPassword]] https://[[attributes.BMC]]/redfish/v1/Systems/system | grep BiosVersion | sed -E 's/.*"(.*)",$/\1/');
```
sinuscosinustan: wouldn't this overwrite the variable regardless of its output and error state?

.firmwareci/workflows/hpe-dl320-g11-main/tests/host-firmware-update.yaml
```diff
@@ -0,0 +1,386 @@
...
+          auth:
+            user: root
+            password: "[[attributes.BMCPassword]]"
+          validate: inventory
+          inventory:
+            rules:
+              - path: /redfish/v1/UpdateService/FirmwareInventory/{{ UpdateTarget }}
+                prop: Updateable
+                op: eq
+                value: "true"
```
sinuscosinustan, line 258: Why should the state change here?

.firmwareci/workflows/hpe-dl320-g11-main/tests/host-firmware-update.yaml
```diff
@@ -0,0 +1,386 @@
...
+        transport: *bmc_ssh
+        parameters:
+          command: "systemctl restart xyz.openbmc_project.GxpChif.service"
+
+  - name: Host on, OS Booting
+    steps:
+      - cmd: shell
+        name: Power on host
+        transport:
+          proto: local
```
sinuscosinustan, line 51:
```suggestion
        transport: &local
          proto: local
```

.firmwareci/workflows/hpe-dl320-g11-main/tests/host-firmware-update.yaml
```diff
@@ -0,0 +1,386 @@
...
+        name: Wait for host OS
+        options:
+          timeout: 3m
+        parameters:
+          host: "[[attributes.Host]]"
+
+      - cmd: shell
+        name: Get BiosVersion
+        transport:
+          proto: local
```
sinuscosinustan:
```suggestion
        transport: *local
```

.firmwareci/workflows/hpe-dl320-g11-main/tests/host-firmware-update.yaml
```diff
@@ -0,0 +1,386 @@
...
+            - regex: \d+\.\d+
+
+  - name: Try update with wrong model (DL360) version
+    steps:
+      - cmd: shell
+        name: Get update endpoint
+        options:
+          timeout: 10s
+        transport:
+          proto: local
```
sinuscosinustan: same here

.firmwareci/workflows/hpe-dl320-g11-main/tests/host-firmware-update.yaml
```diff
@@ -0,0 +1,386 @@
...
+          expect:
+            - regex: 'UEFIPrimary_\d+'
+          variable: UpdateTarget
+
+      - cmd: redfish
+        name: Is updateable?
+        options:
+          timeout: 10s
+        transport:
+          proto: local
```
sinuscosinustan: same here

.firmwareci/workflows/hpe-dl320-g11-main/tests/host-firmware-update.yaml
```diff
@@ -0,0 +1,386 @@
...
+                prop: Updateable
+                op: eq
+                value: "true"
+
+      - cmd: shell
+        name: Trigger update
+        options:
+          timeout: 5m
+        transport:
+          proto: local
```
sinuscosinustan: same here

.firmwareci/workflows/hpe-dl320-g11-main/tests/host-firmware-update.yaml
```diff
@@ -0,0 +1,386 @@
...
+          command: |
+            Task=$(echo {{ UpdateTriggerOutput }} | grep -Eo "/redfish/v1/TaskService/Tasks/[0-9]+" );
+            TaskState="Running";
+            while [ "$TaskState" = "Running" ]; do
+              sleep 2;
+              TaskState=$(curl -sk --user root:[[attributes.BMCPassword]] https://[[attributes.BMC]]${Task} | grep TaskState | sed -E 's/.*"(.*)",$/\1/');
+            done;
+            echo "$TaskState"
+          expect:
+            - regex: Cancelled|Exception
```
sinuscosinustan, line 169: please document why Cancelled and Exception are expected. Exception sounds what it should be (and cancelled only because of the bug)

.firmwareci/workflows/hpe-dl320-g11-main/tests/host-firmware-update.yaml
```diff
@@ -0,0 +1,386 @@
...
+            BIOSVER="null";
+            while [ "$BIOSVER" = "null" ]; do
+              sleep 2;
+              BIOSVER=$(curl -sk --user root:[[attributes.BMCPassword]] https://[[attributes.BMC]]/redfish/v1/Systems/system | grep BiosVersion | sed -E 's/.*"(.*)",$/\1/');
+            done;
+            echo "$BIOSVER"
+          expect:
+            - regex: 2\.92
+
+# on-error: continue    # Ensure all cleanup steps run
```
sinuscosinustan: remove

.firmwareci/workflows/hpe-dl320-g11-main/tests/host-firmware-update.yaml
```diff
@@ -0,0 +1,386 @@
...
+
+  - cmd: dutctl
+    name: Power off DUT
+    parameters:
+      agent: "[[attributes.Agent]]"
+      device: "[[attributes.Device]]"
+      command: power
+      args: ["off"]
+
+  - cmd: dutctl
```
sinuscosinustan, line 378: this shouldn't be only in post-stage but also in pre-stage. The reason is that we can't ensure that the BIOS SPI may got partially overwritten by another (maybe failed) test

sinuscosinustan (changes requested):

.firmwareci/workflows/hpe-dl320-g11-main/tests/host-firmware-update.yaml
haeckle: But shouldn't the copy be performed in case the storage get's updated for whatever reason? Or should it be even in the DUT pre-stage, together with the EV "reset"?

.firmwareci/workflows/hpe-dl320-g11-main/tests/host-firmware-update.yaml
haeckle: yes, good idea. originally wanted to do it under defaults, but there you can only define one alias per entity

.firmwareci/workflows/hpe-dl320-g11-main/tests/host-firmware-update.yaml
haeckle: you're right. did encounter "Cancelled" on a test machine first, but didn't investigate if it was for another reason. here it probably should only be "Exception" for the expected state, will look it up to make sure

.firmwareci/workflows/hpe-dl320-g11-main/tests/host-firmware-update.yaml
haeckle: it only overrides the "null" state and then leaves the loop or do I miss something?
the reason for this loop was initially to not need another wait until the value got propagated. but as I later decided to even wait for the host to boot up there is no reason to make it that complex. will change it to a Redfish inventory check.

.firmwareci/workflows/hpe-dl320-g11-main/tests/host-firmware-update.yaml
haeckle: it shouldn't, but isn't that what a test is for? but semanticly it could be in the stage before. or do you think its superfluous?

.firmwareci/workflows/hpe-dl320-g11-main/tests/host-firmware-update.yaml
haeckle: ok, so should I (also) add it to the DUT pre-stage?

.firmwareci/workflows/hpe-dl320-g11-main/tests/host-firmware-update.yaml
haeckle: Ok, I looked through the code and tried to understand:
When an image for another model is uploaded it gets correctly parsed as PLDM but it fails to provide the matching identifier [1] so the event "UpdateNotApplicable" gets created [2] and the activation is set to "Invalid" [3] which leads to the TaskState "Exception" [4]
On the other hand "Cancelled" is set by the task manager when the timeout is reached (or it is otherwise aborted) [5] which is the case when the flashing with flashcp takes to long [6]
Conclusion: we only really expect "Exception", but take also "Cancelled" as otherwise the next stage wouldn't run. The correct error gets also be captured with the following step.
Will document it in the test description.

[1] https://github.com/openbmc/phosphor-bmc-code-mgmt/blob/dad5ae57eb8b89805d71bfa311da307c563b936d/common/src/device.cpp#L66
[2] https://github.com/openbmc/phosphor-bmc-code-mgmt/blob/dad5ae57eb8b89805d71bfa311da307c563b936d/common/src/device.cpp#L74
[3] https://github.com/openbmc/phosphor-bmc-code-mgmt/blob/dad5ae57eb8b89805d71bfa311da307c563b936d/common/src/device.cpp#L111
[4] https://github.com/openbmc/bmcweb/blob/b2fc97958faaea5230a676f775bf5ea090a464c8/redfish-core/lib/update_service.hpp#L185
[5] https://github.com/openbmc/bmcweb/blob/b2fc97958faaea5230a676f775bf5ea090a464c8/redfish-core/lib/task.hpp#L242
[6] https://gerrit.openbmc.org/c/openbmc/phosphor-bmc-code-mgmt/+/92373

.firmwareci/workflows/hpe-dl320-g11-main/tests/host-firmware-update.yaml
sinuscosinustan: yes

.firmwareci/workflows/hpe-dl320-g11-main/tests/host-firmware-update.yaml
sinuscosinustan: True. I somehow thought that we write it in `pre-stage` but we obv don't do it.

Then I'd say move it to the `pre-stage` and make sure that the BMC runs before / finished its booting. Can be done by checking if the BMC state is `Ready` (we currently have a static sleep of 45 seconds that should be removed).

.firmwareci/workflows/hpe-dl320-g11-main/tests/host-firmware-update.yaml
sinuscosinustan: true, then let's keep it.

.firmwareci/duts/dut-hpe-dl320/pre.yaml
```diff
@@ -50,12 +58,35 @@ pre-stage:
...
+    parameters:
+      source: "[[storage.dl320g11_nvram]]"
+      destination: /tmp
+      recursive: true
+
+  - cmd: shell
+    name: Copy EVs to CHIF path
+    transport: *bmc_ssh
+    parameters:
+      command: "cp /tmp/[[storage.dl320g11_nvram.evs]] /var/lib/chif/evs.dat"
```
haeckle: On the CI the tests including [BIOS](https://app.firmware-ci.com/canopy/jobs/01M0SAHHWVV1EQEQ71D7B66RD0?projectID=01KKGTVRZ3XK1ZDSAETKB1624Y) [reset](https://app.firmware-ci.com/canopy/jobs/01M0SAHHXWF2MP8QZ5GX8F36AW?projectID=01KKGTVRZ3XK1ZDSAETKB1624Y) failing caused by timeouts. While it is expected to take a [while](https://github.com/canopybmc/canopybmc/blob/3b50f7bfe680f2a45725ff0ba7ba74297d460fb6/.firmwareci/workflows/hpe-dl320-g11-ci/tests/bios-config.yaml#L76) for this reset to be completed with this copy of EVs before even a timeout of 30min [seems not enough](https://app.firmware-ci.com/canopy/jobs/01M0SGG089THBEXTPSQRP2S2RN?projectID=01KKGTVRZ3XK1ZDSAETKB1624Y). @sinuscosinustan do you have any idea about this?

sinuscosinustan: No clue. If possible please check it by retrying one of the failing tests and check the KVM via the Web UI

haeckle: ok, that's what I already did and it really took ages for some steps ("BMC Embedded Health Initialization" and "SMBIOS Table Initialization" where for sure both on the lengthy side). Just before it timed out after the 30min I set this was the last screen to show up:
kvm-screenshot-2026-08-24T09-49-57-949Z
Should I try it with an even larger timeout?

haeckle:
> I think the test(s) should be managed in a way that we have a general feedback on the "good" case in the CI and all further (advanced) testing for error conditions go into the long running workflow.

👍️ So I will move this test to [hpe-dl320-g11-ci](https://github.com/canopybmc/canopybmc/tree/main/.firmwareci/workflows/hpe-dl320-g11-ci) and put the incorporated test for the wrong model update on his own test in hpe-dl320-g11-main

> > Current version: https://app.firmware-ci.com/canopy/jobs/01M000YKNRBYN3XVDTWYTVPPZ1?projectID=01KKGTVRZ3XK1ZDSAETKB1624Y
> > while working on it I asked myself and now you, @sinuscosinustan and @mynetz, should we also include further tests, like:
> > 
> > * not a valid pldm image → "VerificationFailed"
> 
> Yes. Invalid input is always a good test case.
> 
> This would be an extended test.

👍️
 
> > * "@Redfish.OperationApplyTime":"OnReset" → "ResetRequired" + manual reset
> 
> This means the update is delayed until a reset happens? Reset of host or reset of BMC?

Reset of the host. If it's "Immediate" it gets rebooted directly after flashing and if [I understand it correctly](https://github.com/openbmc/phosphor-dbus-interfaces/tree/master/yaml/xyz/openbmc_project/Software#image-apply-time) this would be postponed to the next reboot.

> This would be an extended test.

👍️
 
> > * downgrade → "UpdateSuccessful"?
> 
> Is that somth we check/support? I think BIOS downgrades are a security issue. The question is, does the platform care? Could there be reasons for a downgrade? Usually, in case security is higher rated than "safety" this also means the system would need to update both partitions in sequence. In case "safety" is rated higher one of the partitions may stay on the last versions in case you need to fall back for $reasons.
> 
> Not sure what the default case for HPE is or what our default should be.
> 
> I think we should have a issue/feature request for this to discuss this at a later point. Both, general behavior and test-case.

I open up an issue for discussion.

> I did not look to close in the test until now. Are some of the API related tests generic enough that we could (re-)use them for other platforms in the future?

In general it should be be reusable. maybe with some minor adaptions e.g. for the target prefix or expected version formatting

.firmwareci/duts/dut-hpe-dl320/pre.yaml
sinuscosinustan: oh ffs... I may know what the issue is here. A larger timeout will not help in this case.

I'll look into it this week.

haeckle:
@sinuscosinustan would it makes sense to revert [c9ce67c](https://github.com/canopybmc/canopybmc/pull/253/commits/c9ce67c4ef4fc09655f73a4ac6f4199a0f6bb9e6) for now so that the other tests (especially the failing ones including a BIOS reset) are unaffected and we can merge this one?

.firmwareci/workflows/hpe-dl320-g11-ci/tests/host-firmware-update.yaml
```diff
@@ -0,0 +1,207 @@
...
+          timeout: 10m
+          on-error: continue
+        parameters:
+          host: "[[attributes.Host]]"
+
+      - cmd: redfish
+        name: Approve updated BiosVersion
+        options:
+          timeout: 10s
+        transport: &local
```
pmaione: Transport redefinition. Should be `*local`

haeckle: done. also defined it on an earlier step

.firmwareci/workflows/hpe-dl320-g11-ci/tests/host-firmware-update.yaml
```diff
@@ -0,0 +1,205 @@
...
+          auth:
+            user: root
+            password: "[[attributes.BMCPassword]]"
+          validate: inventory
+          inventory:
+            rules:
+              - path: /redfish/v1/Systems/system
+                prop: BiosVersion
+                op: matches
+                value: 2\.92
```
pmaione: Use a template variable to store the bios version instead of hard coding on test step.
```
variables:
  UpdateEndpointPrefix: UEFIPrimary
  ExpectedVersion: "2.92"
...
```
then,
```
 - path: /redfish/v1/Systems/system
   prop: BiosVersion
   op: matches
   value: '{{ ExpectedVersion }}'
```

.firmwareci/workflows/hpe-dl320-g11-ci/tests/host-firmware-update.yaml
```diff
@@ -0,0 +1,205 @@
...
+            password: "[[attributes.BMCPassword]]"
+          validate: inventory
+          inventory:
+            rules:
+              - path: /redfish/v1/Systems/system
+                prop: BiosVersion
+                op: matches
+                value: \d+\.\d+
+
+  - name: Firmware update to 2.92
```
pmaione: Point to the template variable? 
Ex.: Firmware update to `expected version`

.firmwareci/workflows/hpe-dl320-g11-ci/tests/host-firmware-update.yaml
haeckle: sure. I even also made the suggestion in https://github.com/canopybmc/canopybmc/issues/261#issuecomment-5524780804

.firmwareci/workflows/hpe-dl320-g11-ci/tests/host-firmware-update.yaml
haeckle: I would have liked to reuse the variable here. Unfortunately it's not supported by FirmwareCI so I went with a generic "to newer version"

.firmwareci/duts/dut-hpe-dl320/pre.yaml
haeckle: just for reference: this might be caused by #92 and a workaround is implemented with #274

.firmwareci/workflows/hpe-dl320-g11-ci/tests/host-firmware-update.yaml
```diff
@@ -0,0 +1,225 @@
...
+  transport: &bmc_ssh
+    proto: ssh
+    options:
+      host: "[[attributes.BMC]]"
+      user: root
+      password: "[[attributes.BMCPassword]]"
+
+variables:
+  UpdateEndpointPrefix: UEFIPrimary
+  BiosVersionBefore: 2.72 # golden image version, should be lower then update
```
sinuscosinustan: Maybe we can define the BIOS version in the `dl320g11_bios_update` storage instead and link `BiosVersionBefore` to the storage var? Would make it easier in case we ever update the golden image.

haeckle: Let's extract the string directly from the image, so we don't have to bother with this at all

sinuscosinustan (approved):
lgtm. Many thanks!

PR #254, opened 2026-08-07
Make exprtk optional in phosphor-virtual-sensor and add a total_power sensor

pmaione:
Adding `phosphor-virtual-sensor` to the image overflowed rofs by 64 KiB, so the total system power sensor could not ship.

The first commit patches the daemon to make `exprtk` a meson feature option and disables it, dropping roughly 1MB of text. `exprtk` is only needed for the Expression field of the static JSON configuration; sensors configured over D-Bus by entity-manager select their calculation by interface name and never parse an expression.  The commit also replaces the upstream `virtual_sensor_config.json` with one that only sets `Desc.Config` to D-Bus, since the daemon reads entity-manager configs only when the static configuration carries that entry.

The second commit exposes `total_power` on all twelve Gen11 baseboards as a Sum over `PINPUT_PSU1` and `PINPUT_PSU2.` NaN inputs are dropped before summing, so an absent second supply falls out and the sensor reports the remaining one.

rofs lands at 24,981,504 bytes against the 25,214,976 limit, leaving 228 KiB. The daemon adds 102,400 bytes compressed, against roughly 392 KiB with exprtk.

Verified on DL365; total_power has been read on hardware. 

NOTE: `MaxValue` is a clamp, not a threshold, and is set to a uniform 2000 W because no per-platform supply ratings exist in the tree (this value can be discussed and adjusted).


sinuscosinustan (approved):
I was hoping that we can keep exprtk / replace it with something more lightweight, or reduce the compiled features (e.g. vector functions), but apparently the used exprtk variant doesn't have certain macros to disable the majority of bloated features.

Alex has suggested to take a look on https://github.com/codeplea/tinyexpr , maybe this is better suited here (tho it's C only and has no CPP bindings)

PR #255, opened 2026-08-11
Harden IPMI DCMI tests and minor User and Sensor test fixes.

pmaione:
Most DCMI steps carried on-error: continue and accepted any output, so they passed without proving anything. Temperature commands now query the baseboard entity (0x42) instead of inlet (0x40), which has no sensor on this platform, and assert the response bytes, which requires the entity IDs added to the Entity Manager board configs. Asset Tag and power limit become set/get/restore round-trips, the latter setting 800W with exception action No Action, activating it, confirming Get Power Limit reports 800W, and restoring the original state. Get Power Limit outside the round-trip accepts success or completion code 0x80, which DCMI returns when no limit is set, and fails on anything else. The Management Controller ID string is asserted by value, and a duplicate asset tag alias step is removed.

Also fixed a missing step to set user privilege on User Management test and fixed a expected regex for Get Sensor Type in Sensor Events test.

### DIFF
--- a/.firmwareci/workflows/hpe-dl320-g11-ipmi/tests/11-dcmi.yaml
+++ b/.firmwareci/workflows/hpe-dl320-g11-ipmi/tests/11-dcmi.yaml
@@ -1,10 +1,16 @@
 name: "IPMI: DCMI Group Extension Commands (NetFn 0x2c, Group 0xDC)"
 description: >-
   Verify DCMI 1.5 commands (group extension 0xDC). Get DCMI Capability Info,
-  Get Management Controller ID String, Get Temperature Reading, and Get Power
-  Reading are expected to work. Get DCMI Sensor Info (0x07), Get Asset Tag
-  (0x06), and Get Power Limit (0x03) carry on-error: continue as support
-  varies across firmware builds.
+  Get Management Controller ID String, Get DCMI Sensor Info, Get Temperature
+  Reading, and Get Power Reading are expected to work. Temperature commands
+  query the baseboard entity (0x42), backed by GXP Temp; this platform has no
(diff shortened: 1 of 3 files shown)


.firmwareci/workflows/hpe-dl320-g11-ipmi/tests/11-dcmi.yaml
```diff
@@ -27,61 +29,151 @@ stages:
...
+                | head -1 | tr -d '\r\n' > /tmp/fwci_asset_tag_orig
+            elif printf '%s\n' "$OUT" | grep -q 'Parameter out of range'; then
+              : > /tmp/fwci_asset_tag_orig
+            else
+              rm -f /tmp/fwci_asset_tag_orig
+              printf '%s\n' "$OUT"
+              echo "FAIL: could not read current asset tag"
+              exit 1
+            fi
+            echo "saved_asset_tag=[$(cat /tmp/fwci_asset_tag_orig)]"
```
sinuscosinustan: fwci has variables for this ^^

https://docs.firmware-ci.com/references/6_templating_and_variables/index.html#variables

Example (shamelessly copied from #253):
```yaml
      - cmd: shell
        name: Get update endpoint
        options:
          timeout: 30s
        transport:
          proto: local
        parameters:
          command: |
            curl -sk --user root:[[attributes.BMCPassword]] \
              https://[[attributes.BMC]]/redfish/v1/UpdateService/FirmwareInventory \
              | grep -Eo UEFIPrimary_[0-9]+
          expect:
            - regex: 'UEFIPrimary_\d+'
          variable: UpdateTarget
```

sinuscosinustan (approved):

PR #258, opened 2026-08-17
firmwareci: hpe-dl320: continue if nvram flash failed in post-stage

sinuscosinustan:
If the NVRAM can not be flashed, then it should still continue to the point where it powers the system off.

Noticed in the referenced FWCI job.

Link: https://app.firmware-ci.com/canopy/jobs/01M003J5WY4VAX1029EFGH7Q65
Fixes: fd7b1201 ("firmwareci: hpe-dl320: power off via Redfish before turning the DUT off")


haeckle (approved):
LGTM

PR #262, opened 2026-08-21
meta-canopy: validate entity-manager configuration

pointbazaar:
set the PACKAGECONFIG `validate-json` to prevent misconfiguration at build time.


sinuscosinustan:
I can already say that this will break because there is fan config stuff in there that isn't valid :D

